PR #21, opened 2020-12-01
Create PluginsLoader and associated unit tests.

mslukebo:
Refactor assembly processing logic to indicate if all
assemblies found were processed successfully.

Add a DiscoveryStarted method to the IExtensionTypeObserver
interface, which gets called at the start of a round of
processing assemblies.


src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
```diff
@@ -0,0 +1,255 @@
...
+                    {
+                        failed.Add(dir);
+                        continue;
+                    }
+                }
+
+                // TODO: this does redundant work re-finalizing custom data sources
+                // loaded by previous calls to this method. The extension repository
+                // should get refactored to avoid this redundant work.
+                this.extensionRepository.FinalizeDataExtensions();
```
jmaxson-ms, line 159: if we're allowing plug-ins to be loaded at different times, then the extension repository and related code should be updated accordingly. the scenario that I'm thinking of is that a plug-in that has failed to load previously because of missing dependencies might work now. it's been a while, but I believe this could be solved fairly easily by just creating a new DataExtensionDependencyState in each DataExtensionReference where the DataExtensionDependencyState.Availability value is not Undertermined or Available. as an optimization, we might be able to just reset if Availability is MissingRequirement or MissingIndirectRequirement, but we'd need to double check that.

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
```diff
@@ -0,0 +1,255 @@
...
+        /// <remarks>
+        ///     A directory from <paramref name="directories"/> will be included in <paramref name="failed"/>
+        ///     if _any_ of the plugins inside the directory fail to load. See remarks on <see cref="TryLoadPlugin(string)"/>
+        ///     for more information.
+        ///     <para/>
+        ///     This method does not report an error if a different version of one of the custom data
+        ///     sources being loaded has already been loaded by either the current or a previous call to <see cref="TryLoadPlugin(string)"/>
+        ///     or <see cref="TryLoadPlugins(IEnumerable{string}, out IList{string})"/>.
+        /// </remarks>
+        public bool TryLoadPlugins(IEnumerable<string> directories, out IList<string> failed)
```
jmaxson-ms, line 140: >TryLoadPlugins [](start = 20, length = 14)

does it make sense to add an Async version of these calls?

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
```diff
@@ -0,0 +1,255 @@
...
+    {
+        private readonly object mutex = new object();
+
+        private readonly AssemblyExtensionDiscovery extensionDiscovery;
+
+        private readonly IPlugInCatalog catalog;
+
+        private readonly IDataExtensionRepositoryBuilder extensionRepository;
+
+        private readonly ICollection<IPluginsConsumer> subscribers;
```
jmaxson-ms, line 56: >ICollection [](start = 25, length = 11)

because this is private and it's enumerated, would it be better to just use ConcurrentSet here for performance?

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
```diff
@@ -0,0 +1,255 @@
...
+            }
+        }
+
+        /// <summary>
+        ///     Unregisters an <see cref="IPluginsConsumer"/> from plugins loaded in the future.
+        /// </summary>
+        /// <param name="consumer"></param>
+        public void Unsubscribe(IPluginsConsumer consumer)
+        {
+            this.subscribers.Remove(consumer);
```
jmaxson-ms, line 201: >Remove [](start = 29, length = 6)

since we already have a mutex, I'm wondering if we should stop using ConcurrentSet and *only* use the lock. I know this is a pretty extreme example, but imagine a multithreaded caller that calls Subscribe and Unsubscribe on multiple threads. subscribe starts enumerating existing plug-ins, unsubscribe executes on a different thread, then the subscriber is added. I realize that using only the lock doesn't solve all of cases with this, as unsubscribe could run *first*, but in the case described before, if we're only using a lock, then unsubscribe would wait for subscribe to finish before executing.

also, might be worth returning a bool from unsubscribe that indicates whether the subscriber was removed. if nothing else, this might provide an indication that something didn't work as expected for the develop.

i'm not saying that we must make this change, but i believe it's worth considering.

mslukebo: In the case you listed above, the observed behavior would be as if subscribe happened entirely before unsubscribe, which is a valid ordering. The class only guaranteed calls are executed in a linearizable order, and the unsubscribe only guaranteed they do not observe _future_ loaded plugins. Plugins loaded concurrently (i.e. before Unsubscribe returns) or before the call are fine for them to see. 

I agree that we should return a bool indicating if they were successfully unsubscribed, since I can see that being an edge case if the caller is subscribing concurrently (or if they never subscribed at all).

mslukebo: In the case you listed above, the observed behavior would be as if subscribe happened entirely before unsubscribe, which is a valid ordering. The class only guaranteed calls are executed in a linearizable order, and the unsubscribe only guaranteed they do not observe _future_ loaded plugins. Plugins loaded concurrently (i.e. before Unsubscribe returns) or before the call are fine for them to see. 

I agree that we should return a bool indicating if they were successfully unsubscribed, since I can see that being an edge case if the caller is subscribing concurrently (or if they never subscribed at all). 

EDIT: I missed that last part about the actual subscribe being added after unsubscribe returns. In that case it would actually appear as if unsubscribe finished entirely before subscribe, even though the consumer observes the plugins before the collection is modified. The bool would mitigate this issue, but I see the value in using a lock here to prevent that confusion.

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
```diff
@@ -0,0 +1,255 @@
...
+                // Legacy: this plugin was bundled with WPA. It has its binaries inside of one folder
+                // named after the plugin itself.
+                pluginName = parentFolder.Name;
+                pluginVersion = null;
+            }
+            else
+            {
+                // Unknown file system schema. Do not attempt to make assumptions about the plugin's name.
+                // This could happen when loading plugins through un-managed file system schemas (such as the
+                // -addsearchdir option for WPA), or if the calling application does not store
```
jmaxson-ms, line 232: I wonder if we should look into the assembly metadata to find name/version if it's available - or something similar. Not suggesting that should be done in this PR, but maybe we add a todo here?

mslukebo: We could, but that would only provide information about the custom data source's assembly, which may be different than whatever plugin is being loaded. Versions would perhaps even not match. For example, maybe v1 of a plugin uses a CDS v1.5, but then in v2 of the same plugin the author decides v1.5 of the CDS is too buggy and rolls back to include v1.4. If we look in the assembly for this CDS, we'd say "yes this plugin is version 1.4," which may lead people to think it's an earlier version of the plugin.

Also, the CustomDataSourceReference includes its own version and name, so doing this probing isn't necessary since that data is already available.

src/Microsoft.Performance.SDK.Runtime/Discovery/AssemblyExtensionDiscovery.cs
```diff
@@ -191,14 +216,16 @@ public void ProcessAssemblies(IEnumerable<string> directoryPaths)
             }
 
             watch.Stop();
-            Console.Error.WriteLine("Loaded all in {0}", watch.Elapsed);
+            Console.Error.WriteLine("Loaded {0} in {1}", allLoaded ? "all" : "some", watch.Elapsed);
```
jmaxson-ms, line 219: since you're already touching this, would you mind changing watch.Elapsed to watch.ElapsedMilliseconds and then add a 'ms' unit to the string?

Elapsed provides a TimeSpan, and it's output doesn't strike me as particularly intuitive.

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
jmaxson-ms: Oh, so if CDSR already has a version, what do we expect this to be used for?

---
In reply to: [542650493](https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/21#discussion_r542650493) [](ancestors = 542650493)

jmaxson-ms: Oh, nevermind! I didn't realize this was a private method.

---
In reply to: [542673430](https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/21#discussion_r542673430) [](ancestors = 542673430,542650493)

mslukebo: It is private, but the nulls eventually get passed to each consumer's OnCustomDataSourceLoaded.

The nulls signify to the consumer that no plugin information was found for the custom data source being loaded. It's up to the consumers to decide what to do with it then.

In WPA, there will be a default plugin named "Unknown" with no version that enumerates all of the custom data sources loaded with no plugin names.

jmaxson-ms: So are you suggesting we fall back to what's in the CDSR?

---
In reply to: [542677512](https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/21#discussion_r542677512) [](ancestors = 542677512)

mslukebo: No. In WPA, each plugin will be displayed as a list of custom data sources that it provides (since a plugin is just a collection of CDSs). When CDSs are loaded without plugin names/versions, WPA will say "here's all the custom data sources I saw loaded without a plugin," just to let the user know what's currently available. So WPA isn't falling back, it's just not grouping the data sources in a meaningful way.

Either way, we shouldn't fall back on any names/versions in this class, since that'd confuse consumers :)

jmaxson-ms (approved):
:shipit:

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
mslukebo: Is this something that should be done in this PR? This class is primarily a wrapper around the assembly processing, so I'm not sure those changes would make sense for this PR.

FWIW, prior to this class, the SDK "allows" plugins to be loaded at different times; it just doesn't make it easy to do so.

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
mslukebo: I will add some async wrappers for convenience

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
mslukebo: Made this a HashSet since I'm moving to using the mutex for unsubscribe

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
mslukebo: I created #25 as a TODO to deal with this. For reference, the concern of a plugin failing to load initially but then working later because dependencies become available may not be something we even want to deal with. Ideally a plugin includes all of its dependencies that it needs for its custom data sources to work. This scenario really could only happen if plugins are loaded through something like `-addsearchdir` but the user forgot to copy its dependencies to the folder passed in, and then a plugin later miraciously loads the dependency the original one needed. It's possible, but raises the question of how much we want to support loading plugins like that.

src/Microsoft.Performance.SDK.Runtime.NetCoreApp.Tests/Microsoft.Performance.SDK.Runtime.NetCoreApp.Tests.csproj
```diff
@@ -0,0 +1,39 @@
...
+    <Compile Remove="Plugins\MockCustomDataSources\ValidFolderSchema\InstalledPlugins\1.0.0\ValidA\3.0.0\ValidSchemaA2.cs" />
+  </ItemGroup>
+	
+
+  <ItemGroup>
+    <None Include="Plugins\MockCustomDataSources\ValidFolderSchema\InstalledPlugins\1.0.0\ValidA\3.0.0\ValidSchemaA2.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+	<ProjectReference Include="..\Microsoft.Performance.SDK.Runtime.NetCoreApp\Microsoft.Performance.SDK.Runtime.NetCoreApp.csproj" />
```
zachnew-msft, line 29: nit: element indentation should be consistent.

zachnew-msft (dismissed):

zachnew-msft (approved):

jmaxson-ms (approved):

PR #24, opened 2020-12-15
Add CI GitHub workflows

mslukebo:
This PR also bumps NerdBank.GitVersioning to 3.3.37, which is required to allow `dotnet build` to work on GitHub Actions.


mslukebo:
Currently I set up one Action that runs the build and then tests. I created a single badge in the README that indicates if this Action completed succesfully (i.e. builds and all tests pass) for the develop branch.

Two questions we should have are:
1) Do we want badges for main, develop, or both?
2) Do we want two badges where one displays "Build Succeeds" (i.e. an Action that only runs dotnet build) and another that displays "Tests Pass" (i.e. an Action that both builds and runs), or just the combined one I created?

mslukebo:
This PR now contains workflows for the following situations:
1) push to main -> build main
2) push to main -> build main and run tests
3) push to develop -> build develop
4) push to develop -> build develop and run tests
5) pull request into develop or main -> build PR branch and run tests

The last workflow is for CI (and is the one running for this PR). The first four are strictly so we can have different badges for all scenarios. The text inside a badge comes from the workflow's name, so we need separate workflows for main/develop.

Github actions do not support any form of nesting/templating of actions, so unfortunately the workflows are copy-pasted for the most part :(

trgibeau (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
+name: 'PR CI'
+
+on:
+  pull_request:
+    branches: [ develop, main ]
```
zachnew-msft, line 5: We don't need develop anymore

.github/workflows/develop_build_status.yml
```diff
@@ -0,0 +1,29 @@
+name: 'develop build status'
```
zachnew-msft, line 1: We can remove develop specific collateral.

.github/workflows/develop_tests_status.yml
```diff
@@ -0,0 +1,31 @@
+name: 'develop tests status'
```
zachnew-msft, line 1: See comment on develop status.

zachnew-msft (changes requested):

.github/workflows/develop_tests_status.yml
mslukebo: Removed this file

.github/workflows/develop_build_status.yml
mslukebo: Removed this file

.github/workflows/ci.yml
mslukebo: Removed

zachnew-msft (approved):

trgibeau (approved):

PR #51, opened 2021-03-03
Make SetApplicationEnvironmentCore virtual

### LINKED ISSUE #42: Make CustomDataSourceBase.SetApplicationEnvironmentCore virtual
This method is currently abstract. The original implementation of CustomDataSourceBase.SetApplicationEnvironment didn't save the IApplicationEnvironment value anywhere, and so pretty much every derived implementation needed to set it themselves in this method. But now that the value is stored in the ApplicationEnvironment property, most derived classes have nothing to do in SetApplicationEnvironmentCore and shouldn't have to implement an empty method.

mslukebo:
Closes #42


jmaxson-ms:
Should we update our samples at the same time? They probably shouldn't implement this method anymore.

mslukebo:
> Should we update our samples at the same time? They probably shouldn't implement this method anymore.

Our samples use the NuGet feed, so we can't update them until we release a new public version. They're already out of date due to interfaces changes made for 0.109.0. Once we release a new version I'll make a GitHub issue to update samples + docs.

(This further reinforces that we need a preview build feed 😢)

jmaxson-ms (approved):

PR #54, opened 2021-03-08
User/zachnew/disposable

zachnew-msft:
Disposable Extensions

__Overview__

This change introduces the ability for data extensions to be disposable. When the Engine is disposed, all extensions will be cleaned up in an order determined by their dependency graph.

__Usage__

Users may add the `IDisposable` interface to types implementing the following interfaces:
- `ICustomDataSource`
- `ICustomDataProcessor`
- `ISourceDataCooker`
- `ICompositeDataCooker`
- `IDataProcessor`

When this interface is present, then the implementation will be disposed when the engine is disposing. This makes disposal 'opt-in' for users without forcing `IDisposable` onto implementations that may not need it.

Example:
````
using (var engine = Engine.Create())
{
    // ...
} // all cookers, data sources, etc. cleaned up here
````

Users are not required to implement IDisposable. If IDisposable is not present, then no special cleanup logic is performed.

Due to the relationship between CustomDataSources and CustomDataProcessors, an additional method has been added to the `ICustomDataSource` interface: DisposeProcessor. This method is invoked on the Custom Data Source right before a Custom Data Processor is to be cleaned up. This gives the Custom Data Source an opportunity to clean up any state it might have related to the processor. `CustomDataSourceBase` provides a default implementation of this method that does nothing.

__Implementation__

The `IDisposable` interface has been plumbed through all *Reference types and all Repository types, as well as the `Engine`. The `ExtensionRoot` class has been added as the entry point to all loaded plugins and extensions to provide a common point for accessing and cleaning. Upon disposal of the `ExtensionRoot`, a dependency graph is created linking every created item to their dependencies and dependents. The graph is then walked in dependent order, that is, the walk starts with all items that have no dependents, disposes them, and then proceeds to their dependencies, until everything is disposed. Then everything is dropped by either being "nulled out" or collections cleared to signal to the garbage collector that the memory may be reclaimed.

The repositories implement disposal as a failsafe; however, the only way to get the *complete* graph is from the `ExtensionRoot` as it has access to *both* plugins and extensions.

A separate 'Release' path has also been added; this is for scenarios where an application may want to clean up resources without requiring a reload of all of the plugins and extensions; an example would be a UI application that wants to have one instance of the Catalog / ExtensionRepository, but allow for opening and closing different `sessions`.

__Testing__

Tests have been added for *all* types that are implementing `IDisposable` as part of this work. Tests have also been added for the dependency graph and dependency walking.

__Breaking Changes__

The following breaking changes are present in the SDK:

- void DisposeProcessor(ICustomDataProcessor processor) has been added to the `ICustomDataSource` interface.
- Microsoft.PerformanceToolkit.Engine is now `IDisposable`

### CODE REVIEW

File: src/Microsoft.Performance.SDK/Processing/ICustomDataSource.cs [hunk 1]
```diff
@@ -132,5 +132,22 @@ public interface ICustomDataSource
...
+        ///     Provides a method for this Custom Data Source to do any
+        ///     special cleanup operations for processors created by
+        ///     this instance, if applicable.
+        ///     <para />
+        ///     It is guaranteed that the instance passed to this method
+        ///     was created by this instance via call to one of the following:
+        ///     <list type="bullet">
+        ///         <item><see cref="CreateProcessor(IDataSource, IProcessorEnvironment, ProcessorOptions)"/></item>
+        ///         <item><see cref="CreateProcessor(IEnumerable{IDataSource}, IProcessorEnvironment, ProcessorOptions)"/></item>
+        ///     </list>
```
jmaxson-ms, line 146: Is this necessary? "guaranteed" is strong language and if we add additional methods, it might be easy to miss updating this... (and then our guarantee is meaningless and the world will crumble into dust). maybe we should just leave off the method list?

zachnew-msft: That's a good point, I'll remove the list.

src/Microsoft.Performance.SDK.Runtime/AssemblyTypeReference.cs
```diff
@@ -55,25 +64,64 @@ protected AssemblyTypeReference(AssemblyTypeReference<TDerived> other)
...
+        }
+
+        /// <summary>
+        ///     Finalizes an instance of the <see cref="AssemblyTypeReference{TDerived}"/>
+        ///     class.
+        /// </summary>
+        ~AssemblyTypeReference()
+        {
+            this.Dispose(false);
         }
```
jmaxson-ms, line 79: My understanding after reviewing CLR via C# is that adding a finalizer is intended for use when handling a native resource. Adding a finalizer means that the object has exists for an extra generation in the GC, as do any of its references. Since most of these classes don't directly own native resources, I don't believe that adding a finalizer helps, and it actually hurts.

zachnew-msft: You're right, we don't need it.

src/Microsoft.Performance.SDK.Runtime/AssemblyTypeReferenceWithInstance.cs
```diff
@@ -42,13 +62,62 @@ protected AssemblyTypeReferenceWithInstance(Type type)
...
+            private set
+            {
+                this.ThrowIfDisposed();
+                this.instance = value;
+            }
+        }
+
+        /// <summary>
+        ///     Disposes the <see cref="Instance"/> referenced by this instance 
+        ///     without disposing this instance. The <see cref="Instance"/> is
```
jmaxson-ms, line 100: wording here feels a bit ambiguous. perhaps replace "instance" (not the property) with "object"?

src/Microsoft.Performance.SDK.Runtime/ObjectExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+        public static void TryDispose(this object self)
+        {
+            (self as IDisposable)?.SafeDispose();
+        }
+
+        /// <summary>
+        ///     Disposes the given instance of <see cref="IDisposable"/>
+        ///     and enforces the contract that <see cref="IDisposable.Dispose"/>
+        ///     should not throw by catch any <see cref="Exception"/>s that
+        ///     occur.
```
jmaxson-ms, line 33: this only works when using this extension - the actual implementation itself should be sure never to throw. this feels like a false sense of security

src/Microsoft.Performance.SDK.Runtime/AssemblyTypeReference.cs
```diff
@@ -15,9 +15,16 @@ namespace Microsoft.Performance.SDK.Runtime
     /// </typeparam>
     public abstract class AssemblyTypeReference<TDerived>
         : IEquatable<TDerived>,
-          ICloneable<TDerived>
+          ICloneable<TDerived>,
+          IDisposable
```
jmaxson-ms, line 19: >IDisposable [](start = 10, length = 11)

Why add IDisposable to this class?

src/Microsoft.Performance.SDK.Runtime/AssemblyTypeReferenceWithInstance.cs
```diff
@@ -42,13 +62,62 @@ protected AssemblyTypeReferenceWithInstance(Type type)
...
+        ///     then reinitialized.
+        /// </summary>
+        /// <exception cref="System.ObjectDisposedException">
+        ///     This instance is disposed.
+        /// </exception>
+        public void Release()
```
jmaxson-ms, line 106: >public void Release() [](start = 8, length = 21)

We talked about this offline and we don't think we need it.

src/Microsoft.Performance.SDK.Runtime/AssemblyTypeReferenceWithInstance.cs
zachnew-msft: After speaking more offline, we realized this code isn't needed right now, so I am just removing it.

src/Microsoft.Performance.SDK.Runtime/ObjectExtensions.cs
zachnew-msft: This is meant more to be used in the runtime to avoid having try { } catch { } around every call to dispose to avoid an errant extension from blocking cleanup of other things.

src/Microsoft.Performance.SDK.Runtime/AssemblyTypeReference.cs
zachnew-msft: I will add documentation.

src/Microsoft.Performance.SDK.Runtime/ErrorCodes.cs
```diff
@@ -136,6 +136,82 @@ public sealed class ErrorCodes
...
+        /// <summary>
+        ///     A dependency cycle has been detected between extensions.
+        /// </summary>
+        public static ErrorCodes EXTENSION_DependencyCycle = new ErrorCodes(
+            40001,
+            "EXTENSION_DEPENDENCY_CYCLE",
+            "A dependency cycle has been detected between extensions.");
+
+        /// <summary>
+        ///     A dependency cycle has been detected between extensions.
```
jmaxson-ms, line 152: > A dependency cycle has been detected between extensions. [](start = 15, length = 57)

wrong error - copy/paste

src/Microsoft.Performance.SDK.Runtime/ExtensionRoot.cs
```diff
@@ -0,0 +1,189 @@
...
+    public sealed class ExtensionRoot
+        : IPlugInCatalog,
+          IDataExtensionRepositoryBuilder
+    {
+        private DependencyDag dependencyGraph;
+
+        private bool isDisposed;
+
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="ExtensionException"/>
```
jmaxson-ms, line 29: >ExtensionException [](start = 61, length = 18)

haha, i hope not

src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
```diff
@@ -139,38 +269,144 @@ private sealed class CustomDataSourceReferenceImpl
...
+                catch
+                {
+                    try
+                    {
+                        this.Instance.DisposeProcessor(processor);
+                    }
+                    finally
+                    {
+                        processor.TryDispose();
+                    }
```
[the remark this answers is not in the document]
zachnew-msft: That extension method does a null check.

src/Microsoft.Performance.SDK.Runtime/ErrorCodes.cs
zachnew-msft: Thanks!

src/Microsoft.Performance.SDK.Runtime/ExtensionRoot.cs
zachnew-msft: Thanks!

mslukebo:
Should we add this to the advanced documentation as a feature you can add?

src/Microsoft.Performance.SDK.Runtime/ExtensionRoot.cs
```diff
@@ -0,0 +1,189 @@
...
+            if (disposing)
+            {
+                try
+                {
+                    var dag = DependencyDag.Create(this.Catalog, this.Repository);
+                    dag.DependentWalk(
+                        x => x.Target.Match(
+                            r => r.SafeDispose(),
+                            r => r.SafeDispose()));
+                }
```
jmaxson-ms, line 170: It seems like we should catch/eat any exception here - given that this is in a Dispose method

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/DataExtensionAvailabilityError.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Performance.SDK.Runtime.Extensibility.DataExtensions
+{
+    class DataExtensionAvailabilityError
```
jmaxson-ms, line 7: what's up with this?

zachnew-msft:
> Should we add this to the advanced documentation as a feature you can add?

Yes, I can add a doc for it.

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/DataExtensionAvailabilityError.cs
zachnew-msft: I forgot to delete it. Thanks!

src/Microsoft.Performance.SDK.Runtime/ExtensionRoot.cs
zachnew-msft: I'll update it.

jmaxson-ms:
>     internal class DataProcessorReference

would you mind making this sealed?

---
Refers to: src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/DataProcessors/DataProcessorReference.cs:18 in 6c75313. [](commit_id = 6c753139e662b0d9b27087bcd15bd45094b2c1d3, deletion_comment = False)

zachnew-msft:
> > Should we add this to the advanced documentation as a feature you can add?
> 
> Yes, I can add a doc for it.

I have added documentation.

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/DataCookers/CompositeDataCookerReference.cs
```diff
@@ -1,19 +1,89 @@
...
 using System.Diagnostics;
+using Microsoft.Performance.SDK.Extensibility;
+using Microsoft.Performance.SDK.Extensibility.DataCooking;
 
 namespace Microsoft.Performance.SDK.Runtime.Extensibility.DataExtensions.DataCookers
 {
+    /// <summary>
+    ///     Represents a reference to a composite data cooker.
+    /// </summary>
     internal class CompositeDataCookerReference
```
jmaxson-ms, line 14: Would you mind sealing this class? If you're ok with that, then 'instanceLock' could be made private (it probably should have been already).

jmaxson-ms:
>     internal class SourceDataCookerReference

another class that I think can be sealed, if you wouldn't mind

---
Refers to: src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/DataCookers/SourceDataCookerReference.cs:20 in 0732b90. [](commit_id = 0732b90256fa7cf213a9d9816e2e9226861e92b8, deletion_comment = False)

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Dependency/DependencyDag.cs
```diff
@@ -0,0 +1,797 @@
...
+        public static DependencyDag Create(
+            IPlugInCatalog catalog,
+            IDataExtensionRepository repository)
+        {
+            Guard.NotNull(catalog, nameof(catalog));
+            Guard.NotNull(repository, nameof(repository));
+
+
+            if (!catalog.IsLoaded)
+            {
```
jmaxson-ms, line 84: nit: extra line

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Dependency/DependencyDag.cs
```diff
@@ -0,0 +1,797 @@
...
+        /// </exception>
+        /// <exception cref="System.ArgumentNullException">
+        ///     <paramref name="catalog"/> is <c>null</c>.
+        ///     - or -
+        ///     <paramref name="repository"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="System.InvalidOperationException">
+        ///     A dependency cycle exists between two or more
+        ///     componenets in the graph.
+        /// </exception>
```
jmaxson-ms, line 74: Probably need to add:
        /// </exception>

since catalog.IsLoaded can throw this. I hate that this needs to be manually bubbled up. i feel like we're going to be missing this exception everywhere now

zachnew-msft: Thanks, I'll add it.

zachnew-msft: After looking at it more, I'll catch any ObjectDisposedExceptions and rethrow as InvalidOperation exceptions as it is not valid to compute the graph from disposed repositories.

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Dependency/DependencyDag.cs
```diff
@@ -0,0 +1,797 @@
...
+            //    / \                             |
+            //   d   |  will be simplified to     d
+            //    \  |                            |
+            //      d2                            d2
+            //
+            // This is because n needs d and d2, but d also needs d2, so
+            // fundamentally n only needs d.
+            //
+
+            var roots = allNodes.Where(x => x.Dependents.Count == 0).ToSet();
```
jmaxson-ms, line 161: >roots [](start = 16, length = 5)

not a big deal, but since we already have all the roots, could we just pass this into the constructor rather than recalculating?

zachnew-msft:
> > ```
> > internal class DataProcessorReference
> > ```
> 
> would you mind making this sealed?
> 
> Refers to: src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/DataProcessors/DataProcessorReference.cs:18 in [6c75313](https://github.com/microsoft/microsoft-performance-toolkit-sdk/commit/6c753139e662b0d9b27087bcd15bd45094b2c1d3). [](commit_id = [6c75313](https://github.com/microsoft/microsoft-performance-toolkit-sdk/commit/6c753139e662b0d9b27087bcd15bd45094b2c1d3), deletion_comment = False)

Yes, I have now sealed it.

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Dependency/DependencyDag.cs
zachnew-msft: Yes, that's a good point. Thanks!

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Dependency/DependencyDag.cs
```diff
@@ -0,0 +1,797 @@
...
+                        .Cast<IDataExtensionReference>())
+                .Concat(
+                    der.RequiredDataProcessors
+                        .Select(reps.GetDataProcessorReference)
+                        .Cast<IDataExtensionReference>())
+                .Select(Reference.Create)
+                .ToSet();
+        }
+
+        private static void SimplifyTransitiveDependencies(Node node)
```
jmaxson-ms, line 234: >SimplifyTransitiveDependencies [](start = 28, length = 30)

I'm wondering if this would be more efficient as a recursive method. probably not a big deal, just seems like we're doing a bunch of extra work this way.

jmaxson-ms: oh, i didn't look closely enough at the Node structure. this might be fine. :)

---
In reply to: [591759896](https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/54#discussion_r591759896) [](ancestors = 591759896)

zachnew-msft: I don't know how big graphs can get, so I did it iteratively as I didn't want to potentially blow the stack.

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
```diff
@@ -45,16 +44,17 @@ namespace Microsoft.Performance.SDK.Runtime.NetCoreApp.Plugins
     ///     with their <see cref="Subscribe(IPluginsConsumer)"/> call.
     /// </summary>
     public class PluginsLoader
```
jmaxson-ms, line 46: >PluginsLoader [](start = 17, length = 13)

maybe seal this class as well?

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -199,13 +333,17 @@ public static Engine Create()
...
         ///     This instance has already been processed.
         /// </exception>
+        /// <exception cref="ObjectDisposedException">
+        ///     This instance is disposed.
+        /// </exception>
         /// <exception cref="UnsupportedDataSourceException">
         ///     <paramref name="dataSource"/> cannot be processed by any
         ///     discovered extensions.
-        /// </exception>
+        /// </exception>=
```
jmaxson-ms, line 342: >= [](start = 24, length = 1)

typo?

jmaxson-ms (approved):
:shipit:

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
zachnew-msft: Yes, thanks!

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Plugins/PluginsLoader.cs
zachnew-msft: Sure

PR #56, opened 2021-03-10
Add Table support to Engine

trgibeau:
Enabled the ability to access tables from the engine

- Added Try/EnableTable(TableDescriptor tableDescriptor)
-- Allows callers to enable the table before calling Process()
- Added IsTableDataAvailable(TableDescriptor tableDescriptor)
-- Allows callers to check if the table has data after calling Process()
- Added Try/BuildTable(TableDescriptor tableDescriptor)
-- Allows callers to build a table after calling Process
- Added Unit and Functional tests


src/Microsoft.Performance.Toolkit.Engine/ITableResult.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Performance.SDK.Processing;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Performance.Toolkit.Engine
+{
+    public interface ITableResult
```
jmaxson-ms, line 10: we'll need to have doc comments on this since it's public :)

src/Microsoft.Performance.Toolkit.Engine.Tests/TestData/EngineBuildTabeTestDtos.cs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Performance.Toolkit.Engine.Tests.TestData
+{
+    [DataContract]
+    public class EngineBuildTableTestSuiteDto
```
anramanath, line 10: What is Dto?

jmaxson-ms: https://en.wikipedia.org/wiki/Data_transfer_object

jmaxson-ms:
Going to have to merge Zack's IDispose change on the Engine.

anramanath (approved):

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -433,6 +451,56 @@ public bool TryEnableCooker(DataCookerPath dataCookerPath)
...
+        /// <summary>
+        ///     Enables the given table for processing when <see cref="Process"/>
+        ///     is called.
+        /// </summary>
+        /// <param name="descriptor">
+        ///     The table to enable.
+        /// </param>
+        public void EnableTable(TableDescriptor descriptor)
+        {
+            this.ThrowIfProcessed();
```
mslukebo, line 463: can we add a <throws> to the docstring

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -433,6 +451,56 @@ public bool TryEnableCooker(DataCookerPath dataCookerPath)
...
+
+            if (!this.TryEnableTable(descriptor))
+            {
+                throw new TableNotFoundException(descriptor);
```
mslukebo, line 467: and one for this

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
```diff
@@ -41,10 +46,12 @@ public sealed class RuntimeExecutionResults
         ///     - or -
         ///     <paramref name="repository"/> is <c>null</c>.
         /// </exception>
+        /// TODO: Going to move RuntimeExecutionResults to internal and expose calls via new interface
```
mslukebo, line 49: can we denote this with a `<remarks>` (or even an `ObsoleteAttribute`)

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
```diff
@@ -215,5 +223,217 @@ public bool TryQueryOutput<T>(DataOutputPath dataOutputPath, out T data)
...
+        ///     Checks if the given <paramref name="tableDescriptor"/> has data available.
+        /// </summary>
+        /// <param name="tableDescriptor">
+        ///     The table to check.
+        /// </param>
+        /// <returns>
+        ///     <c>true</c> the table has data available;
+        ///     <c>false</c> otherwise.
+        ///     <c>null</c> the table does not support checking if data is available, 
+        ///     recommend to call <see cref="BuildTable(TableDescriptor)"/>.
```
mslukebo, line 237: what does this mean

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
```diff
@@ -215,5 +223,217 @@ public bool TryQueryOutput<T>(DataOutputPath dataOutputPath, out T data)
...
+        ///     Builds a table with the given <paramref name="tableDescriptor"/>.
+        /// </summary>
+        /// <param name="tableDescriptor">
+        ///     The table to build.
+        /// </param>
+        /// <returns>
+        ///     The built table.
+        /// </returns>
+        /// <exception cref="TableNotBuiltException">
+        ///     Table cannot be built for the given <paramref name="tableDescriptor"/>.
```
mslukebo, line 264: nit: A table

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
```diff
@@ -215,5 +223,217 @@ public bool TryQueryOutput<T>(DataOutputPath dataOutputPath, out T data)
...
+        /// </exception>
+        public ITableResult BuildTable(TableDescriptor tableDescriptor)
+        {
+            if(TryBuildTable(tableDescriptor, out ITableResult filledTableBuilder))
```
mslukebo, line 268: nit: space after if

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
trgibeau: IsDataAvailable is not required to be implemented. 

Should reword a little?

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
trgibeau: I'm going to do this in a different PR (didn't want it to get too big :))

It will be before we snap 0.109

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
mslukebo: the part I was confused on is "recommend to call <see..."

trgibeau: Ah, wanted to say that the caller should call BuildTable directly without needing to check if it has data

trgibeau:
> Going to have to merge Zack's IDispose change on the Engine.

Thank Jayson, just remerged and added dispose support

@zachnew-msft can you double check that I added dispose in the right places

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -48,6 +50,9 @@ public abstract class Engine
         private List<DataCookerPath> enabledCookers;
         private ReadOnlyCollection<DataCookerPath> enabledCookersRO;
 
+        private readonly List<TableDescriptor> enabledTables;
+        private readonly ReadOnlyCollection<TableDescriptor> enabledTablesRO;
```
jmaxson-ms, line 54: these shouldn't be readonly because they need to be nulled in Dispose

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -269,6 +278,30 @@ private set
             }
         }
 
+        /// <summary>
+        ///     Gets the collection of all enabled tables
+        /// </summary>
+        public IEnumerable<TableDescriptor> EnableTables
```
jmaxson-ms, line 284: >EnableTables [](start = 44, length = 12)

typo: EnabledTables

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -269,6 +278,30 @@ private set
             }
         }
 
+        /// <summary>
+        ///     Gets the collection of all enabled tables
+        /// </summary>
```
jmaxson-ms, line 283: for each of the new methods/properties, this should be added to the doc comments:

        /// </exception>

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -598,6 +631,64 @@ public bool TryEnableCooker(DataCookerPath dataCookerPath)
...
+        /// <param name="descriptor">
+        ///     The table to enable.
+        /// </param>
+        /// <returns>
+        ///     <c>true</c> if the table exists and can be enabled;
+        ///     <c>false</c> otherwise. Note that <c>false</c> is
+        ///     always returned when <see cref="IsProcessed"/> is <c>true</c>.
+        /// </returns>
+        public bool TryEnableTable(TableDescriptor descriptor)
```
jmaxson-ms, line 670: >descriptor [](start = 51, length = 10)

should we Guard.NotNull(descriptor)? if so, we should also add the ArgumentNullException to the doc comments (and bubbled up to EnableTable)

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -863,10 +956,38 @@ private RuntimeExecutionResults ProcessCore()
...
+            var extendedTables = new HashSet<ITableExtensionReference>();
+            var processorTables = new Dictionary<TableDescriptor, ICustomDataProcessor>();
+
+            foreach (var table in this.enabledTables)
+            {
+                foreach (var executor in executors.Where(x => x.Context.CustomDataSource.AvailableTables.Contains(table)))
+                {
+                    executor.Processor.EnableTable(table);
+                    processorTables.Add(table, executor.Processor);
+                }
```
jmaxson-ms, line 969: I'm wondering if this should be part of an "else" block below the if block below. Given the a table is uniquely identified by a Guid, it would be a bug to be in both sets.

jmaxson-ms:
>         ///     Gets any non-fatal errors that occured during engine

would you mind fixing this typo while you're modifying this file?

---
Refers to: src/Microsoft.Performance.Toolkit.Engine/Engine.cs:259 in 02d70c2. [](commit_id = 02d70c2b6f25ba990013c9e0bc8bcfec59413e4d, deletion_comment = False)

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
```diff
@@ -41,10 +46,12 @@ public sealed class RuntimeExecutionResults
         ///     - or -
         ///     <paramref name="repository"/> is <c>null</c>.
         /// </exception>
+        /// TODO: Going to move RuntimeExecutionResults to internal and expose calls via new interface
         public RuntimeExecutionResults(
             ICookedDataRetrieval cookedDataRetrieval,
             IDataExtensionRetrievalFactory retrievalFactory,
-            IDataExtensionRepository repository)
+            IDataExtensionRepository repository,
+            IDictionary<TableDescriptor, ICustomDataProcessor> tableToProcessorMap)
```
jmaxson-ms, line 54: >tableToProcessorMap [](start = 63, length = 19)

Guard.NotNull and update doc comment for ArgumentNullException

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
```diff
@@ -215,5 +223,222 @@ public bool TryQueryOutput<T>(DataOutputPath dataOutputPath, out T data)
...
+                {
+                    var tableRetrieval = this.retrievalFactory.CreateDataRetrievalForTable(tableDescriptor.Guid);
+
+                    result = func(reference, tableRetrieval);                    
+
+                    return true;
+                }
+                catch
+                {
+                    return false;
```
jmaxson-ms, line 330: should we report some error to the user in this case? (same in TryOnProcessor) if we want to , we might have to let the exception bubble up to where we know what the func is actually doing

jmaxson-ms: gah, just realized this is in a Try[Method]. not sure what the right thing to do here is

---
In reply to: [592041314](https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/56#discussion_r592041314) [](ancestors = 592041314)

jmaxson-ms: maybe just have a TryGetOn[Repository|Processor], and not pass in the func?

---
In reply to: [592041488](https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/56#discussion_r592041488) [](ancestors = 592041488,592041314)

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
```diff
@@ -20,6 +24,7 @@ public sealed class RuntimeExecutionResults
         private readonly IDataExtensionRetrievalFactory retrievalFactory;
         private readonly IDataExtensionRepository repository;
         private readonly IEnumerable<DataCookerPath> sourceCookers;
+        private readonly IDictionary<TableDescriptor, ICustomDataProcessor> tableToProcessorMap;
```
jmaxson-ms, line 27: >ICustomDataProcessor [](start = 54, length = 20)

is it possible for this Table to exist in more than one instance of the same ICustomDataProcessor?

src/Microsoft.Performance.Toolkit.Engine/ITableResult.cs
jmaxson-ms: Sorry, one more doc comment on the interface itself

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
trgibeau: This is going to moved to internal in another PR

trgibeau: Added Guard.NotNull

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
jmaxson-ms: An engine is a single session, so this shouldn't be an issue.

---
In reply to: [592042897](https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/56#discussion_r592042897) [](ancestors = 592042897)

trgibeau: nope :)

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -863,10 +967,40 @@ private RuntimeExecutionResults ProcessCore()
...
+                if (this.extensionRoot.TablesById.TryGetValue(table.Guid, out ITableExtensionReference reference))
+                {
+                    extendedTables.Add(reference);
+                }
+                else
+                {
+                    foreach (var executor in executors.Where(x => x.Context.CustomDataSource.AvailableTables.Contains(table)))
+                    {
+                        executor.Processor.EnableTable(table);
+                        processorTables.Add(table, executor.Processor);
```
jmaxson-ms, line 985: based on what I remember @zachnew-msft said today... an executor is tied to a CDS, and a table should not exist in multiple custom data sources, so I think semantically it would make more sense to do a .First() instead of a foreach.

clearly not a functional change, just seems more clear to me what we actually expect to happen if we did it that way

src/Microsoft.Performance.Toolkit.Engine/EngineException.cs
```diff
@@ -189,6 +189,172 @@ public override string ToString()
...
+        ///     class with the given <see cref="TableDescriptor"/>, and the error that is the cause of this error.
+        /// </summary>
+        /// <param name="tableDescriptor">
+        ///     The <see cref="TableDescriptor"/> that was not found.
+        /// </param>
+        /// <param name="inner">
+        ///     The error that is the cause of this error.
+        /// </param>
+        public TableException(TableDescriptor tableDescriptor, Exception inner)
+            : base($"The requested table '{tableDescriptor}' was not found", inner)
```
jmaxson-ms, line 315: >$"The requested table '{tableDescriptor}' was not found" [](start = 19, length = 56)

this should be updated

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
```diff
@@ -215,5 +225,228 @@ public bool TryQueryOutput<T>(DataOutputPath dataOutputPath, out T data)
...
+            }
+            catch (TableException)
+            {
+                filledTableBulder = null;
+                return false;
+            }            
+        }
+
+        // Will perform the Func on the Repository if the table is contained, else false.
+        private bool FuncOnRepositoryIfContained<TResult>(TableDescriptor tableDescriptor, Func<ITableExtensionReference, IDataExtensionRetrieval, TResult> func, out TResult result)
```
jmaxson-ms, line 335: >FuncOnRepositoryIfContained [](start = 21, length = 27)

My personal preference is to avoid types in names - so something like "ExecuteOnRepositoryIfContained"... but not a big deal with way. Feel free to ignore :)

src/Microsoft.Performance.Toolkit.Engine/EngineException.cs
```diff
@@ -189,6 +189,172 @@ public override string ToString()
...
+                .ToString();
+        }
+    }
+
+    /// <summary>
+    ///     Represents the error that occurs when an attempt is made to build a
+    ///     <see cref="ITableResult"/> that with a specified <see cref="TableDescriptor"/>.
+    /// </summary>
+    [Serializable]
+    public class TableException
```
jmaxson-ms, line 280: >TableException [](start = 17, length = 14)

I wonder if we should require a message on this exceptions as it is fairly generic

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
trgibeau: Tweaked it to be "Single" since it should be there

jmaxson-ms (approved):
:shipit:

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
```diff
@@ -215,5 +225,229 @@ public bool TryQueryOutput<T>(DataOutputPath dataOutputPath, out T data)
...
+                }
+
+                if (ExecuteOnProcessorIfContained(tableDescriptor, (processor) => processor.DoesTableHaveData(tableDescriptor), out bool processorIsDataAvail))
+                {
+                    return processorIsDataAvail;
+                }
+            }
+            catch (Exception inner)
+            {
+                throw new TableException($"An excpetion was thrown while calling IsDataAvailable for the {tableDescriptor}.", tableDescriptor, inner);
```
jmaxson-ms, line 265: >excpetion [](start = 46, length = 9)

typo

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
```diff
@@ -215,5 +225,229 @@ public bool TryQueryOutput<T>(DataOutputPath dataOutputPath, out T data)
...
+                {
+                    return processorTableResult;
+                }
+            }
+            catch (Exception inner)
+            {
+                innerException = inner;
+            }
+
+            throw new TableException($"An excpetion was thrown while calling BuildTable for the {tableDescriptor}.", tableDescriptor, innerException);
```
jmaxson-ms, line 305: >excpetion [](start = 42, length = 9)

copy/paste

src/Microsoft.Performance.Toolkit.Engine/EngineException.cs
```diff
@@ -189,6 +189,178 @@ public override string ToString()
...
+        : EngineException
+    {
+        /// <inheritdoc />
+        public TableException() : base() { }
```
jmaxson-ms, line 284: >public [](start = 8, length = 6)

want this private?

trgibeau: I think since its serializable it needs to be public

jmaxson-ms (approved):

jmaxson-ms (approved):

zachnew-msft (approved):

PR #61, opened 2021-03-12
Add migration document for 0.109

zachnew-msft:
Adding a folder for documenting migrations due to breaking changes.


mslukebo:
@zachnew-msft should we add that `CustomDataSourceBase.SetApplicaitonEnvironmentCore` is now virtual, so it can be removed?

zachnew-msft:
> @zachnew-msft should we add that `CustomDataSourceBase.SetApplicaitonEnvironmentCore` is now virtual, so it can be removed?

That's not a breaking change, so I didn't think of it. I can add a note if it would be helpful.

mslukebo:
> > @zachnew-msft should we add that `CustomDataSourceBase.SetApplicaitonEnvironmentCore` is now virtual, so it can be removed?
> 
> That's not a breaking change, so I didn't think of it. I can add a note if it would be helpful.

It might be useful to have a [Breaking Changes] section and a [Optional Changes] section in this. If people are going to be making code changes when upgrading they have one place to look for everything they might want/need to change.

zachnew-msft:
> > > @zachnew-msft should we add that `CustomDataSourceBase.SetApplicaitonEnvironmentCore` is now virtual, so it can be removed?
> > 
> > 
> > That's not a breaking change, so I didn't think of it. I can add a note if it would be helpful.
> 
> It might be useful to have a [Breaking Changes] section and a [Suggested Changes] section in this. If people are going to be making code changes when upgrading they have one place to look for everything they might want/need to change.

That's a great idea. Thanks!

documentation/Migration/Update-0.109.md
```diff
@@ -0,0 +1,97 @@
...
+
+# Suggested Changes
+
+The following are changes that are not required, but may be useful to you.
+
+## CustomDataSourceBase
+
+The method `SetApplicationEnvironmentCore` is now virtual, so you no longer have
+to override it if you do not want to. An `ApplicationEnvironment` property is
+now available on the base class that you may reference in your `CustomDataSources.`
```
mslukebo, line 97: nit: `CustomDataSources.` -> `CustomDataSource`.

documentation/Migration/Update-0.109.md
```diff
@@ -0,0 +1,97 @@
...
+}
+catch (UnsupportedDataSourceException)
+{
+    // ...
+}
+````
+
+## IDataSource
+
+Anywhere you are calling `IDataSource.GetUri()` can be simply updated to `IDataSource.Uri`
```
mslukebo, line 87: I would prefer these be kept solely as lists of changes, at least for each intro sentence. This should start with "`IDataSource.GetUri()` has been changed to a read-only property: `IDataSource.Uri`.

documentation/Migration/Update-0.109.md
zachnew-msft: Thanks!

documentation/Migration/Update-0.109.md
zachnew-msft: I'll update it

jmaxson-ms (dismissed):

jmaxson-ms (approved):

trgibeau (approved):

Issue #65, opened 2021-03-30
Discovery errors are not displaying their details when printed


zachnew-msft:
SDK Version: 0.109

Repro:

Try to load a plugin that references an invalid version of the SDK (e.g. try to load a plugin that references 0.107 using 0.109)
Capture the out error parameter from the plugin loader.

Expected:
When the error is printed, all of the details are exposed.

Actual:
Details, Target, etc. are missing.

PR #66, opened 2021-03-30
Update ErrorInfo formatting.

### LINKED ISSUE #64: ErrorInfo with collection properties does not print information correctly.
SDK Version: 0.109

Repro:

Subclass ErrorInfo
Add a property of Type 'List<string>' to your new type
Call ToString on an instance of your type

Expected:
The elements of the list property get printed

Actual:
System.Collections.Generic.List`1[string] gets printed

zachnew-msft:
Fixes #64 and #65

Update ErrorInfo printing logic.

Some error properties, such as lists, were not printing properly. Rather than
printing all of their items, they were simply printing the type name, which is
not useful.

Some other internal errors had overridden ToString unnecessarily, which led to
a loss of information.

This change rectifies these issues so that printing is consistent and useful.


jmaxson-ms (approved):
:shipit:

trgibeau (approved):

PR #68, opened 2021-04-06
Take develop to main

zachnew-msft:
RI develop to main.


zachnew-msft:
I need to fix the merge conflicts.

documentation/Components.md
```diff
@@ -0,0 +1,128 @@
+# Components
+
+This file outlines and provides a brief description of the various components that make up the SDK.
```
anramanath, line 3: Can we add a line here saying that the page is under construction. I feel bad for readers who might see that the expanders and then only after clicking it will they notice it's not filled in.

anramanath (approved):

documentation/Components.md
zachnew-msft: We will add it in a new PR.

jmaxson-ms (approved):
:shipit:

firoj0:
Online advertising marketing social media all brands ads

PR #72, opened 2021-04-13
Do not enumerate types from non-extension DLLs.

zachnew-msft:
Not all DLLs reference the SDK. If a DLL does not reference the SDK, then
there is no way that said DLL can contain a plugin or extension. This change
will cause the extension discovery to skip over any DLLs that do not reference
the SDK when doing Type enumeration.

After validating that the DLL is acceptable according to the validator, it is
loaded and passed to ProcessAssembly. Here, we add a check for an SDK reference
and return early if the reference is not present.

Tests have been added for this scenario.


src/Microsoft.Performance.Testing/FakeAssembly.cs
```diff
@@ -12,23 +11,35 @@ public sealed class FakeAssembly
...
+        public override AssemblyName GetName()
         {
-            Assert.IsNotNull(this.TypesToReturn);
+            return this.AssemblyName;
+        }
 
-            return this.TypesToReturn;
+        public override AssemblyName[] GetReferencedAssemblies()
+        {
+            return this.ReferencedAssemblies ?? Array.Empty<AssemblyName>();
```
jmaxson-ms, line 37: >?? Array.Empty<AssemblyName>(); [](start = 45, length = 31)

why is this needed?

jmaxson-ms (approved):
:shipit:

src/Microsoft.Performance.Testing/FakeAssembly.cs
jmaxson-ms: asking for my own benefit. if it's not needed, i'm not suggestion you need to change it. just curious if there's a path I'm not thinking of.

trgibeau (approved):

PR #73, opened 2021-04-13
Update CustomDataSourceReference Supports to filter correctly.

zachnew-msft:
CustomDataSourceReference was blindly passing datasources directly to the underlying
CustomDataSource without checking if it was even remotely plausible that the CDS could
accept the data source.

The supports method has been updated to interrogate the data source attributes before
asking the Custom Data Source.

The CustomDataSourceReference class had to be refactored to make this testable; all
tests were updated to use the CustomDataSourceReference class directly and still pass without
changing behavior. This also has the side effect of making the CustomDataSourceReference
class easier to understand.


src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
```diff
@@ -42,13 +61,46 @@ protected CustomDataSourceReference(Type type)
...
+        /// </param>
         /// <exception cref="System.ArgumentNullException">
         ///     <paramref name="type"/> is <c>null</c>.
         /// </exception>
-        protected CustomDataSourceReference(Type type, Func<ICustomDataSource> instanceFactory)
+        public CustomDataSourceReference(
+            Type type,
+            Func<ICustomDataSource> instanceFactory,
+            CustomDataSourceAttribute metadata,
+            HashSet<DataSourceAttribute> dataSourceAttributes)
```
jmaxson-ms, line 74: copy/paste the doc comments from the other constructor

src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
```diff
@@ -61,36 +113,72 @@ protected CustomDataSourceReference(Type type, Func<ICustomDataSource> instanceF
...
             : base(other.Type)
         {
+            this.dataSourceAttributes = other.dataSourceAttributes;
+
+            this.guid = other.guid;
+            this.name = other.name;
+            this.description = other.description;
+            this.commandLineOptionsRO = other.commandLineOptionsRO;
+
+            this.isDisposed = other.isDisposed;
```
jmaxson-ms, line 126: > this.isDisposed = other.isDisposed; [](start = 11, length = 36)

if other.isDisposed is true, we should never get this far and I'm not sure it makes sense to apply it to this instance anyway... since this instance's disposed state doesn't depend on other. this doesn't look right to me

src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
```diff
@@ -61,36 +113,72 @@ protected CustomDataSourceReference(Type type, Func<ICustomDataSource> instanceF
         /// <exception cref="System.ObjectDisposedException">
         ///     <paramref name="other"/> is disposed.
         /// </exception>
-        protected CustomDataSourceReference(CustomDataSourceReference other)
+        public CustomDataSourceReference(CustomDataSourceReference other)
```
jmaxson-ms, line 116: do we want to copy createdprocessor[ro]?

src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
```diff
@@ -14,9 +14,19 @@ namespace Microsoft.Performance.SDK.Runtime
     /// <summary>
     ///     Creates a <see cref="AssemblyTypeReferenceWithInstance{T, Derived}"/> where T is <see cref="ICustomDataSource"/> and Derived is <see cref="CustomDataSourceReference"/>.
     /// </summary>
-    public abstract class CustomDataSourceReference
+    public sealed class CustomDataSourceReference
```
mslukebo, line 17: this is a breaking change; should we document it in the migrations/changes docs

jmaxson-ms: This is in the runtime. I don't believe that we note breaking changes for the runtime.

---
In reply to: [613291810](https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/73#discussion_r613291810) [](ancestors = 613291810)

jmaxson-ms: We recommend developers use the engine to access data, rather than directly accessing the runtime.

---
In reply to: [613293042](https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/73#discussion_r613293042) [](ancestors = 613293042,613291810)

src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
```diff
@@ -207,314 +343,133 @@ protected CustomDataSourceReference(CustomDataSourceReference other)
...
-                        catch
-                        {
-                        }
+            base.Dispose(disposing);
+            this.isDisposed = true;
+        }
 
-                        p.TryDispose();
-                    }
+        private void InitializeThis(
```
mslukebo, line 456: since we can't do a ctor for this initialization, should we add a bool `this.isInitialized` that gets set to `true` in this method? Then anywhere we do `ThrowIfDisposed` we do a `ThowIfNotInitialized`. Just to make sure we actually call this

src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
mslukebo: I agree; if we were using `other`'s public accessors it'd throw if it was disposed

src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
```diff
@@ -207,314 +343,133 @@ protected CustomDataSourceReference(CustomDataSourceReference other)
...
-            /// <inheritdoc />
-            public override IReadOnlyCollection<DataSourceAttribute> DataSources
-            {
-                get
-                {
-                    this.ThrowIfDisposed();
-                    return this.dataSourceAttributes;
-                }
-            }
+                hash = ((hash << 5) + hash) ^ this.Name.GetHashCode();
```
mslukebo, line 405: can/should we use HashCodeUtils here?

zachnew-msft: yes, thanks

src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
zachnew-msft: sure. That should probably be an assert as it is a bug in the runtime to not call it, and the caller has no control over that at all.

src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
zachnew-msft: That is a good point, I'll have this just throw if other is disposed.

src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
zachnew-msft: Thanks

src/Microsoft.Performance.SDK.Runtime/CustomDataSourceReference.cs
zachnew-msft: Yes, I will look at it.

jmaxson-ms (approved):

trgibeau (approved):

PR #74, opened 2021-04-15
Expose all extensions from the loader for now.

zachnew-msft:
We should expose all extensions consistently from the runtime to avoid accidentally using different instances of repositories.


jmaxson-ms (approved):
:shipit:

mslukebo (approved):

PR #75, opened 2021-04-19
Generate shared assemblies to include references

jmaxson-ms:
Our static list of shared assemblies only contained the SDK assemblies. The generated list will additionally contain assemblies that are in the reference chain.

### DIFF
--- a/src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Discovery/IsolationAssemblyLoader.cs
+++ b/src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Discovery/IsolationAssemblyLoader.cs
@@ -9,6 +9,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using System.Runtime.Loader;
 using Microsoft.Performance.SDK.Runtime.Discovery;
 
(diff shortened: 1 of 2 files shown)


zachnew-msft (changes requested):

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Discovery/IsolationAssemblyLoader.cs
```diff
@@ -116,13 +108,67 @@ internal static void ClearLoadContexts()
...
+                    continue;
+                }
+
+                fileNamesToAssemblyNames.Add(Path.GetFileName(assembly.Location), assemblyName);
+                processedAssemblies.Add(assemblyName);
+
+                var referencedAssemblies = assembly.GetReferencedAssemblies();
+                foreach (var referencedAssemblyName in referencedAssemblies)
+                {
+                    var referencedAssembly = Assembly.Load(referencedAssemblyName.ToString());
```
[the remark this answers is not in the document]
zachnew-msft: I just realized that I was thinking about this backwards.

zachnew-msft (approved):

src/Microsoft.Performance.SDK.Runtime.Tests/Discovery/IsolatedAssemblyLoadingTests.cs
```diff
@@ -57,6 +57,40 @@ public void Cleanup()
...
+            Assert.IsTrue(assemblyNames.Contains("Microsoft.Performance.SDK.Runtime.NetCoreApp"));
+            Assert.IsTrue(assemblyNames.Contains("Microsoft.Performance.SDK.Runtime"));
+            Assert.IsTrue(assemblyNames.Contains("Microsoft.Performance.SDK"));
+        }
+
+        [TestMethod]
+        [UnitTest]
+        public void AlwaysSharedAssembliesContainReferencedAssemblies()
+        {
+            // Nuget.Versioning is currently referenced, so using it as a test case.
```
mslukebo, line 76: would it make sense to shim GetReferencedAssemblies?

jmaxson-ms: I don't think it's worth it for this considering we'll be using nuget.versioning for semantic versioning for the foreseeable future. If we run into problems, then I can go ahead an over engineer this. :)

mslukebo (approved):
:shipit:

PR #78, opened 2021-06-14
Do metadata inspection per folder.

zachnew-msft:
Similar to how to load each folder into a separate Assembly Load Context, this change updates the metadata validation to also load per folder.

### DIFF
--- a/src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Discovery/SandboxPreloadValidator.cs
+++ b/src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Discovery/SandboxPreloadValidator.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
@@ -20,7 +21,7 @@ namespace Microsoft.Performance.SDK.Runtime.NetCoreApp.Discovery
     {
         private readonly MetadataAssemblyResolver resolver;
         private readonly VersionChecker versionChecker;
-        private readonly MetadataLoadContext loadContext;
+        private readonly ConcurrentDictionary<string, MetadataLoadContext> loadContextsByDirectory;
 
         private bool isDisposed;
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Discovery/SandboxPreloadValidator.cs
```diff
@@ -37,15 +38,19 @@ public sealed class SandboxPreloadValidator
 
             this.resolver = new PathAssemblyResolver(allDlls);
             this.versionChecker = versionChecker;
-            this.loadContext = new MetadataLoadContext(this.resolver);
```
mslukebo, line 40: was this intentional

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Discovery/SandboxPreloadValidator.cs
```diff
@@ -37,15 +38,19 @@ public sealed class SandboxPreloadValidator
...
         public bool IsAssemblyAcceptable(string fullPath, out ErrorInfo error)
         {
+            var loadContext = this.loadContextsByDirectory.GetOrAdd(
+                Path.GetDirectoryName(fullPath),
+                _ => this.CreateLoadContext());
+
             Assembly loaded;
             try
             {
-                loaded = this.loadContext.LoadFromAssemblyPath(fullPath);
```
mslukebo, line 48: was this intentional

mslukebo (approved):

src/Microsoft.Performance.SDK.Runtime.NetCoreApp/Discovery/SandboxPreloadValidator.cs
zachnew-msft: Yes, the load context is now retrieved from the dictionary instead of being a single instance variable.

mslukebo: Completely missed that :)

zachnew-msft (commented):
Placeholder.

trgibeau (approved):

PR #93, opened 2021-07-13
Additional functionality and fixes.

jmaxson-ms:
Additional data retrieval methods and extension methods.
Engine may use multiple plugin paths.
Additional engine creation info.
Fixes XPerf symbols deadlock.
Updated and new tests.

### DIFF
--- a/src/Microsoft.Performance.SDK.Runtime.Tests/Extensibility/CompositeDataCookerReferenceTests.cs
+++ b/src/Microsoft.Performance.SDK.Runtime.Tests/Extensibility/CompositeDataCookerReferenceTests.cs
@@ -121,14 +121,12 @@ namespace Microsoft.Performance.SDK.Runtime.Tests.Extensibility
 
             public bool TryQueryOutput<T>(DataOutputPath identifier, out T result)
             {
-                result = default;
-                return false;
+                throw new NotImplementedException();
             }
 
             public bool TryQueryOutput(DataOutputPath identifier, out object result)
             {
-                result = default;
-                return false;
+                throw new NotImplementedException();
             }
         }
 
--- a/src/Microsoft.Performance.SDK.Runtime.Tests/Extensibility/CompositeDataCookerReferenceTests.cs
+++ b/src/Microsoft.Performance.SDK.Runtime.Tests/Extensibility/CompositeDataCookerReferenceTests.cs
@@ -118,7 +118,20 @@ namespace Microsoft.Performance.SDK.Runtime.Tests.Extensibility
             {
                 throw new NotImplementedException();
             }
+
+            public bool TryQueryOutput<T>(DataOutputPath identifier, out T result)
+            {
+                result = default;
+                return false;
+            }
+
+            public bool TryQueryOutput(DataOutputPath identifier, out object result)
+            {
+                result = default;
+                return false;
+            }
         }
+
         private sealed class FakeSupport
         : IDataExtensionDependencyStateSupport
         {
--- a/src/Microsoft.Performance.SDK.Runtime.Tests/Extensibility/TestClasses/TestSourceDataCooker.cs
+++ b/src/Microsoft.Performance.SDK.Runtime.Tests/Extensibility/TestClasses/TestSourceDataCooker.cs
@@ -33,12 +33,42 @@ namespace Microsoft.Performance.SDK.Runtime.Tests.Extensibility.TestClasses
 
         public T QueryOutput<T>(DataOutputPath identifier)
         {
-            throw new System.NotImplementedException();
+            if (Reflector == null)
+            {
+                throw new NotImplementedException();
+            }
+
+            return Reflector.QueryOutput<T>(identifier);
         }
 
         public object QueryOutput(DataOutputPath identifier)
         {
-            throw new System.NotImplementedException();
+            if (Reflector == null)
+            {
+                throw new NotImplementedException();
+            }
+
+            return Reflector.QueryOutput(identifier);
+        }
+
+        public bool TryQueryOutput<T>(DataOutputPath identifier, out T result)
+        {
+            if (Reflector == null)
+            {
+                throw new NotImplementedException();
+            }
+
+            return Reflector.TryQueryOutput<T>(identifier, out result);
+        }
+
+        public bool TryQueryOutput(DataOutputPath identifier, out object result)
+        {
+            if (Reflector == null)
+            {
+                throw new NotImplementedException();
+            }
+
+            return Reflector.TryQueryOutput(identifier, out result);
         }
 
         public IReadOnlyCollection<DataOutputPath> OutputIdentifiers { get; set; }
@@ -92,5 +122,14 @@ namespace Microsoft.Performance.SDK.Runtime.Tests.Extensibility.TestClasses
         //public int CountOfTestRecordsReceived = 0;
 
         public TestSourceDataCookerContext Context { get; set; }
+
+
+        internal TestCookedDataReflector Reflector { get; private set; }
+
+        internal void SetCookedDataReflector(TestCookedDataReflector reflector)
+        {
+            Reflector = reflector;
+            OutputIdentifiers = reflector.OutputIdentifiers;
+        }
     }
 }
--- a/src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/DataCookers/CompositeDataCookerReference.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/DataCookers/CompositeDataCookerReference.cs
@@ -68,9 +68,8 @@ namespace Microsoft.Performance.SDK.Runtime.Extensibility.DataExtensions.DataCoo
         ///     A <see cref="Type"/> must satisfy the following criteria in order to 
         ///     be eligible as a reference:
         ///     <list type="bullet">
-        ///         <item>must be public.</item>
         ///         <item>must be concrete.</item>
-        ///         <item>must implement ICompositeDataCookerDescriptor somewhere in the inheritance heirarchy (either directly or indirectly.)</item>
+        ///         <item>must implement ICompositeDataCookerDescriptor somewhere in the inheritance hierarchy (either directly or indirectly.)</item>
         ///         <item>must have a public parameterless constructor.</item>
         ///     </list>
         /// </summary>
@@ -92,7 +91,7 @@ namespace Microsoft.Performance.SDK.Runtime.Extensibility.DataExtensions.DataCoo
 
             reference = null;
 
-            if (candidateType.IsPublicAndInstantiatableOfType(typeof(ICompositeDataCookerDescriptor)))
+            if (candidateType.IsInstantiatableOfType(typeof(ICompositeDataCookerDescriptor)))
             {
                 // There must be an empty, public constructor
                 if (candidateType.TryGetEmptyPublicConstructor(out var constructor))
@@ -122,7 +121,7 @@ namespace Microsoft.Performance.SDK.Runtime.Extensibility.DataExtensions.DataCoo
         /// <exception cref="System.ObjectDisposedException">
         ///     This instance is disposed.
         /// </exception>
-        protected ICompositeDataCookerDescriptor Instance
+        private ICompositeDataCookerDescriptor Instance
         {
             get
             {
--- a/src/Microsoft.Performance.SDK.Runtime/Extensibility/FilteredDataRetrieval.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/Extensibility/FilteredDataRetrieval.cs
@@ -35,8 +35,6 @@ namespace Microsoft.Performance.SDK.Runtime.Extensibility
         /// <inheritdoc/>
         public T QueryOutput<T>(DataOutputPath dataOutputPath)
         {
-            Guard.NotNull(dataOutputPath, nameof(dataOutputPath));
-
             if (!string.IsNullOrWhiteSpace(dataOutputPath.SourceParserId))
             {
                 // this is a source cooker, so it already exists
@@ -51,8 +49,6 @@ namespace Microsoft.Performance.SDK.Runtime.Extensibility
         /// <inheritdoc/>
         public object QueryOutput(DataOutputPath dataOutputPath)
         {
-            Guard.NotNull(dataOutputPath, nameof(dataOutputPath));
-
             if (!string.IsNullOrWhiteSpace(dataOutputPath.SourceParserId))
             {
                 // this is a source cooker, so it already exists
@@ -61,6 +57,34 @@ namespace Microsoft.Performance.SDK.Runtime.Extensibility
             else
             {
                 return this.QueryCompositeOutput(dataOutputPath);
+            }
+        }
+
+        /// <inheritdoc/>
+        public bool TryQueryOutput<T>(DataOutputPath dataOutputPath, out T result)
+        {
+            if (!string.IsNullOrWhiteSpace(dataOutputPath.SourceParserId))
+            {
+                // this is a source cooker, so it already exists
+                return this.TryQuerySourceOutput<T>(dataOutputPath, out result);
+            }
+            else
+            {
+                return this.TryQueryCompositeOutput<T>(dataOutputPath, out result);
+            }
+        }
+
+        /// <inheritdoc/>
+        public bool TryQueryOutput(DataOutputPath dataOutputPath, out object result)
+        {
+            if (!string.IsNullOrWhiteSpace(dataOutputPath.SourceParserId))
+            {
+                // this is a source cooker, so it already exists
+                return this.TryQuerySourceOutput(dataOutputPath, out result);
+            }
+            else
+            {
+                return this.TryQueryCompositeOutput(dataOutputPath, out result);
             }
         }
 
@@ -102,6 +126,11 @@ namespace Microsoft.Performance.SDK.Runtime.Extensibility
             }
         }
 
+        private bool TryValidateSourceDataCooker(DataOutputPath dataOutputPath)
+        {
+            return this.extensionDependencies.RequiredSourceDataCookerPaths.Contains(dataOutputPath.CookerPath);
+        }
+
         private IDataCooker ValidateCompositeDataCooker(DataOutputPath dataOutputPath)
         {
             var cookerPath = dataOutputPath.CookerPath;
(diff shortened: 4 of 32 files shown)


src/Microsoft.Performance.SDK.Runtime/Extensibility/SourceDataCookerDependencyProvider.cs
```diff
@@ -33,9 +33,22 @@ public TOut QueryOutput<TOut>(DataOutputPath dataOutputPath)
             return (TOut)QueryOutput(dataOutputPath);
         }
 
+        public bool TryQueryOutput<TOut>(DataOutputPath dataOutputPath, out TOut result)
```
zachnew-msft, line 36: We seem to be using just 'T' on the other methods instead of 'TOut.' Should we use just 'T' here for consistency?

src/Microsoft.Performance.SDK.Runtime/TypeExtensions.cs
```diff
@@ -19,5 +19,17 @@ internal static bool TryGetEmptyPublicConstructor(this Type type, out Constructo
             constructor = null;
             return false;
         }
+
+        internal static bool TryGetEmptyConstructor(this Type type, out ConstructorInfo constructor)
+        {
+            constructor = type.GetConstructor(Type.EmptyTypes);
+            if (constructor != null)
```
zachnew-msft, line 26: we can just return constructor != null instead of the if / else, as if the constructor is null, we set it to null and return false anyway.

src/Microsoft.Performance.SDK/Extensibility/ICookedDataRetrievalExtensions.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// <param name="result">
+        ///     The uniquely identified data.
+        /// </param>
+        /// <exception cref="InvalidOperationException">
+        ///     The specified <paramref name="dataOutputPath"/> was not found.
+        /// </exception>
+        /// <exception cref="InvalidCastException">
+        ///     The value at <paramref name="dataOutputPath"/> cannot be cast to <typeparamref name="T"/>.
+        /// </exception>
+        public static void QueryOutput<T>(
```
zachnew-msft, line 36: Why does this extension exist? It returns void and uses an out parameter. Do we need it?

src/Microsoft.Performance.SDK/Extensibility/ICookedDataRetrievalExtensions.cs
```diff
@@ -0,0 +1,201 @@
...
+        ///     The uniquely identified data.
+        /// </param>
+        /// <exception cref="InvalidOperationException">
+        ///     The specified <paramref name="outputId"/> was not found on the data cooker 
+        ///     <paramref name="cookerPath"/>.
+        /// </exception>
+        /// <exception cref="InvalidCastException">
+        ///     The value at <paramref name="identifier"/> cannot be cast to <typeparamref name="T"/>.
+        /// </exception>
+        public static void QueryOutput<T>(
```
zachnew-msft, line 160: Same as for ````void QueryOutput<T>```` at the top of the file.

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -80,5 +102,23 @@ public string ExtensionDirectory
...
         /// </summary>
         public VersionChecker Versioning { get; internal set; }
+
+        /// <summary>
+        ///     Gets or sets the name of the runtime on which the application is built.
+        /// </summary>
+        /// <remarks>
+        ///     Defauls to "Microsoft.Performance.Toolkit.Engine".
+        /// </remarks>
+        public string RuntimeName { get; set; } = "Microsoft.Performance.Toolkit.Engine";
```
zachnew-msft, line 112: Do we want to add any validation against null / whitespace here?

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -14,53 +17,72 @@ namespace Microsoft.Performance.Toolkit.Engine
...
                 {
-                    throw new InvalidExtensionDirectoryException(value);
+                    throw new InvalidExtensionDirectoryException(directory);
                 }
 
-                this.extensionDirectory = dirInfo.FullName;
+                directories.Add(dirInfo.FullName);
             }
+
+            this.ExtensionDirectories = directories.AsReadOnly();
```
zachnew-msft, line 77: Does it make sense to dedupe this list here?

zachnew-msft (changes requested):

src/Microsoft.Performance.SDK.Runtime/Extensibility/SourceDataCookerDependencyProvider.cs
jmaxson-ms: This class is already using T as a type parameter, so for these methods, we use TOut instead.

src/Microsoft.Performance.SDK/Extensibility/ICookedDataRetrievalExtensions.cs
jmaxson-ms: It makes it so that you don't need to specify the type T when making the call. For example, this works...

QueryOutput(path, out bool result);

src/Microsoft.Performance.SDK/Extensibility/ICookedDataRetrievalExtensions.cs
jmaxson-ms: Same answer.

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
jmaxson-ms: Zack and I spoke offline. This makes sense, but DirectoryInfo doesn't implement IEquatable or similar and making this work across platforms would be challenging. We're just going to leave this for now.

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
jmaxson-ms: The engine handles this. I think null/empty should be ok here.

                string runtimeName = !string.IsNullOrWhiteSpace(createInfo.RuntimeName)
                    ? createInfo.RuntimeName
                    : "Microsoft.Performance.Toolkit.Engine";

zachnew-msft (dismissed):

zachnew-msft (approved):

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -80,5 +102,23 @@ public string ExtensionDirectory
...
+
+        /// <summary>
+        ///     Gets or sets the application name.
+        /// </summary>
+        public string ApplicationName { get; set; }
+
+        /// <summary>
+        ///     Gets or sets a means of logging information.
+        /// </summary>
+        public ILogger Logger { get; set; }
```
zachnew-msft, line 122: I think it would make more sense to have this be a logger factory, i.e. Func<Type, ILogger> so that the user is in control of whether the same logger gets reused across all data sources or not. As this appears to be implemented in the engine, if this property is set, then it is used for all Custom Data Sources, but if it is not set, then a new Logger is created for each Custom Data Source. Using a factory function provides that same flexibility but still allows the caller to do something else if desiered.

jmaxson-ms: Updated.

src/Microsoft.Performance.SDK.Runtime.Tests/Extensibility/CompositeDataCookerReferenceTests.cs
```diff
@@ -118,7 +118,20 @@ public object QueryOutput(DataOutputPath identifier)
             {
                 throw new NotImplementedException();
             }
+
+            public bool TryQueryOutput<T>(DataOutputPath identifier, out T result)
+            {
+                result = default;
```
mslukebo, line 124: is there a reason these don't just also `throw new NotImplementedException`

src/Microsoft.Performance.SDK.Runtime.Tests/Extensibility/TestClasses/TestSourceDataCooker.cs
```diff
@@ -33,12 +33,42 @@ public TestSourceDataCooker(TestSourceDataCookerContext context)
 
         public T QueryOutput<T>(DataOutputPath identifier)
         {
-            throw new System.NotImplementedException();
+            if (Reflector == null)
```
mslukebo, line 36: Should we have a separate text fixture for the reflection tests?

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/DataCookers/CompositeDataCookerReference.cs
```diff
@@ -92,7 +91,7 @@ private CompositeDataCookerReference(Type type)
 
             reference = null;
 
-            if (candidateType.IsPublicAndInstantiatableOfType(typeof(ICompositeDataCookerDescriptor)))
+            if (candidateType.IsInstantiatableOfType(typeof(ICompositeDataCookerDescriptor)))
```
mslukebo, line 94: what's the reason for this change?

src/Microsoft.Performance.SDK.Runtime/Extensibility/FilteredDataRetrieval.cs
```diff
@@ -64,6 +60,34 @@ public object QueryOutput(DataOutputPath dataOutputPath)
             }
         }
 
+        /// <inheritdoc/>
+        public bool TryQueryOutput<T>(DataOutputPath dataOutputPath, out T result)
+        {
+            if (!string.IsNullOrWhiteSpace(dataOutputPath.SourceParserId))
```
mslukebo, line 66: we should probably get the dataoutput path PR in and then check the DataOutputPathType instead

src/Microsoft.Performance.SDK.Runtime/Extensibility/FilteredDataRetrieval.cs
```diff
@@ -102,6 +126,11 @@ private void ValidateSourceDataCooker(DataOutputPath dataOutputPath)
             }
         }
 
+        private bool TryValidateSourceDataCooker(DataOutputPath dataOutputPath)
+        {
+            return this.extensionDependencies.RequiredSourceDataCookerPaths.Contains(dataOutputPath.CookerPath);
```
mslukebo, line 131: do we want to avoid using `.CookerPath`?

zachnew-msft (approved):

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/DataCookers/CompositeDataCookerReference.cs
jmaxson-ms: It's required to support non-public custom data source, which is something we want specifically for DataLayer, but would also be good to have in general.

src/Microsoft.Performance.SDK.Runtime/Extensibility/FilteredDataRetrieval.cs
jmaxson-ms: I don't know, is there?

src/Microsoft.Performance.SDK.Runtime.Tests/Extensibility/CompositeDataCookerReferenceTests.cs
jmaxson-ms: I'm just in the habit of not throwing from Try methods :). I'll add it.

src/Microsoft.Performance.SDK.Runtime.Tests/Extensibility/TestClasses/TestSourceDataCooker.cs
jmaxson-ms: Perhaps, but I don't want to update for this PR as this was already being used.

src/Microsoft.Performance.SDK.Runtime/Extensibility/FilteredDataRetrieval.cs
jmaxson-ms: Whichever is checked in first, the other should update this. Either way, I don't care.

src/Microsoft.Performance.SDK/Extensibility/ICookedDataRetrievalExtensions.cs
```diff
@@ -0,0 +1,202 @@
...
+            DataCookerPath cookerPath,
+            string outputId,
+            out object result)
+        {
+            return retrieval.TryQueryOutput(new DataOutputPath(cookerPath, outputId), out result);
+        }
+
+        /// <summary>
+        ///     Retrieves data by name, typecast to the given type.
+        ///     The caller is coupled to the data extension and expected to know the
```
mslukebo, line 101: what exactly does "The caller is coupled to the data extension" mean? Is this just saying to call this generic method the caller needs to know at compile time what T is? Isn't that implied with a typeparam in the signature?

src/Microsoft.Performance.SDK/Processing/CustomDataProcessorBaseWithSourceParser.cs
```diff
@@ -131,41 +131,103 @@ public void EnableCooker(ISourceDataCookerFactory sourceDataCookerFactory)
...
-        /// <param name="dataOutputPath">Path to the data output</param>
-        /// <returns>Data output of the specified type</returns>
+        /// <inheritdoc/>
+        /// <remarks>
+        ///     Returns data from a source data cooker registered with <see cref="SourceProcessingSession"/>.
+        /// </remarks>
+        /// <exception cref="ArgumentException">
+        ///     <paramref name="dataOutputPath"/> does not target the data processor.
+        /// </exception>
         public TOutput QueryOutput<TOutput>(DataOutputPath dataOutputPath)
```
mslukebo, line 141: why do we even have methods for querying data cookers on a custom data processor? seems like we're blurring boundaries between object functionality

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -805,6 +808,22 @@ protected void ThrowIfProcessed()
...
+        /// </summary>
+        /// <param name="type">
+        ///     The <c>Type</c> for which the logger is created.
+        /// </param>
+        /// <returns>
+        ///     An instance of <see cref="ILogger"/>.
+        /// </returns>
+        protected ILogger CreateLogger(Type type)
+        {
+            Debug.Assert(!this.isDisposed);
```
mslukebo, line 822: is this line redundant with the line below?

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -14,53 +17,72 @@ namespace Microsoft.Performance.Toolkit.Engine
...
+        /// <param name="extensionDirectory">
+        ///     The extension directory from which the runtime instance is to load plugins.
+        /// </param>
+        /// <remarks>
+        ///     If no extension directory is specified, the current directory is used.
+        /// </remarks>
+        /// <exception cref="InvalidExtensionDirectoryException">
+        ///     Thrown if <paramref name="extensionDirectory"/> is an invalid directory path.
+        /// </exception>
+        public EngineCreateInfo(string extensionDirectory = null)
```
mslukebo, line 32: thoughts on a separate parameterless ctor instead of using a default?

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
jmaxson-ms: No, this can make it easier to find while debugging.

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -80,5 +102,26 @@ public string ExtensionDirectory
...
         /// </summary>
         public VersionChecker Versioning { get; internal set; }
+
+        /// <summary>
+        ///     Gets or sets the name of the runtime on which the application is built.
+        /// </summary>
+        /// <remarks>
+        ///     Defaults to "Microsoft.Performance.Toolkit.Engine".
+        /// </remarks>
+        public string RuntimeName { get; set; } = "Microsoft.Performance.Toolkit.Engine";
```
mslukebo, line 112: are we able to grab the assembly name dynamically/at compile time instead of hard-coding this?

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
```diff
@@ -19,6 +18,7 @@ namespace Microsoft.Performance.Toolkit.Engine
     ///     Represents the results of processing data using the <see cref="Engine"/>.
     /// </summary>
     public sealed class RuntimeExecutionResults
+        : ICookedDataRetrieval
```
mslukebo, line 21: was this class already implementing this interface implicitly? I don't see anything added to the class.

Why are "results" able to retrieve data? I feel like that should be happening directly through the engine/runtime, not through its final result. but that might be a bigger change

src/Microsoft.Performance.SDK/Extensibility/ICookedDataRetrievalExtensions.cs
jmaxson-ms: It means that it must know the return type - there can be no compile time type check. And if that return type changes, the caller must be updated.

src/Microsoft.Performance.SDK/Processing/CustomDataProcessorBaseWithSourceParser.cs
jmaxson-ms: This does blur things a bit, you're correct and we can chat offline. But I'm going to resolve this comment and it's not new functionality to this PR.

src/Microsoft.Performance.Toolkit.Engine/EngineExecutionResults.cs
jmaxson-ms: This class was already implementing the interface, yes. Adding the interface allows the interface extensions to work with this class.

The results are how clients of the engine retrieve any data from the output. @zachnew-msft designed it this way, and it seems to work well. But if you have some thoughts on how to improve it, happy to chat.

src/Microsoft.Performance.SDK.Runtime/Extensibility/FilteredDataRetrieval.cs
mslukebo: I was just thinking if we're trying to use the rep of a cooker path less maybe we should be passing around/using the actual DataOutPath objects instead. We'd could just implement IComparable/IEquatable on the paths

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
jmaxson-ms: sure, i'm ok with that.

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
jmaxson-ms: That's probably a good idea.

src/Microsoft.Performance.SDK.Runtime/Extensibility/FilteredDataRetrieval.cs
jmaxson-ms: This isn't a string, it's a DataCookerPath, which does implement IEquatable<>. And We want to compare the cooker paths, not the DataOutputPath, which is why we're using that property.

mslukebo: Ah, my bad. I thought this DataCookerPath.CookerPath which was the string.

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -14,53 +17,87 @@ namespace Microsoft.Performance.Toolkit.Engine
...
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="EngineCreateInfo"/> class.
+        /// </summary>
+        /// <exception cref="ArgumentNullException">
+        ///     Thrown is <paramref name="extensionDirectories"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="InvalidExtensionDirectoryException">
+        ///     Thrown if <paramref name="extensionDirectories"/> contains an invalid directory path.
+        /// </exception>
+        public EngineCreateInfo(IEnumerable<string> extensionDirectories)
```
mslukebo, line 61: missing a <param> in docstring 😃

mslukebo (dismissed):

zachnew-msft (approved):

mslukebo (approved):

zachnew-msft (approved):

PR #113, opened 2021-08-02
User/anramanath/process data element array

### LINKED ISSUE #110: Optimize DataCooker processing path
Update BaseSourceProcessingSession<T, TContext, TKey>.ProcessDataElement to use a for loop rather than a foreach loop, and to use an array rather than an IEnumerable.

anramanath:
Making ProcessDataElement more efficient. fixes #110.


src/Microsoft.Performance.SDK.Runtime/Extensibility/BaseSourceProcessingSession.cs
```diff
@@ -1,16 +1,16 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System;
```
mslukebo, line 4: I think we're deciding to keep System imports first. You can change this behavior in VS Options under Text Editor -> C# -> Advanced

src/Microsoft.Performance.SDK.Runtime/Extensibility/BaseSourceProcessingSession.cs
```diff
@@ -116,17 +116,18 @@ public DataProcessingResult ProcessDataElement(T data, TContext context, Cancell
 
             var key = data.GetKey();
 
-            if (!this.activeCookerRegistry.ContainsKey(key) && !this.receiveAllDataCookers.Any())
+            if (!this.activeCookerRegistry.ContainsKey(key) && this.receiveAllDataCookers.Count == 0)
```
mslukebo, line 119: why this change?

src/Microsoft.Performance.SDK.Runtime/Extensibility/BaseSourceProcessingSession.cs
```diff
@@ -143,17 +144,18 @@ DataProcessingResult ProcessDataCookers(IEnumerable<ISourceDataCooker<T, TContex
...
-                var sourceDataCookers = this.activeCookerRegistry[key];
                 keyedResult = ProcessDataCookers(sourceDataCookers);
                 if (keyedResult == DataProcessingResult.CorruptData)
                 {
                     return keyedResult;
                 }
             }
 
-            var allDataResult = ProcessDataCookers(this.activeReceiveAllDataCookers);
+            var allDataCookers = this.activeReceiveAllDataCookers;
```
mslukebo, line 157: un-needed variable

src/Microsoft.Performance.SDK.Runtime/Extensibility/BaseSourceProcessingSession.cs
```diff
@@ -190,15 +192,16 @@ protected void InitializeForSourceParsing()
...
         }
 
         protected void ActivateCooker(ISourceDataCooker<T, TContext, TKey> sourceDataCooker)
         {
             Guard.NotNull(sourceDataCooker, nameof(sourceDataCooker));
 
             this.activeCookers.Add(sourceDataCooker);
-            if (sourceDataCooker.Options.HasFlag(SourceDataCookerOptions.ReceiveAllDataElements))
+            if (sourceDataCooker.Options.HasFlag(SourceDataCookerOptions.ReceiveAllDataElements) &&
+                !this.activeReceiveAllDataCookers.Contains(sourceDataCooker))
```
mslukebo, line 204: this might be expensive. is this a hot path?

mslukebo (changes requested):

src/Microsoft.Performance.SDK.Runtime/Extensibility/BaseSourceProcessingSession.cs
anramanath: Using .Any() requires an enumerator. Since Count is cached, we can access that faster.

src/Microsoft.Performance.SDK.Runtime/Extensibility/BaseSourceProcessingSession.cs
anramanath: It could be expensive, but should not be a problem as it isn't a hot path.

mslukebo (approved):

zachnew-msft (approved):

mslukebo (approved):

zachnew-msft (approved):

Issue #109, opened 2021-07-23
Engine Message Box implementation needs to handle user prompts sanely


trgibeau:
Right now, the message box class in /Microsoft.Performance.Toolkit.Engine/Engine.cs does not know what to do with a message box that requires user interaction. We need to decide what to do.

trgibeau:
For 1.0: 

- rename GraphicalUserEnvironment bool to "IsInteractive"
- Update Engine to support setting if the caller is interactive

anramanath:
A couple of days of adding and testing.

PR #119, opened 2021-08-16
Feature/engine immediate feedback

zachnew-msft:


documentation/Migration/Update-0.109.md
```diff
@@ -36,11 +36,41 @@ protected override bool IsFileSupportedCore(string path)
...
+engine.AddDataSource(new FileDataSource("test.txt"));
+engine.EnableCooker(cookerPath);
+
+// ... use the engine as desired.
+
+````
+becomes
+````cs
+
+using var dataSources = DataSourceSet.Create();
```
mslukebo, line 58: the usings in this and above don't have corresponding closure scopes. should they?

documentation/Using-the-SDK/Using-the-engine.md
```diff
@@ -0,0 +1,104 @@
...
+The `Engine` is used to programmatically process data. There are three (3)
+major components to be aware of when using the `Engine`:
+- The `PluginSet`
+- The `DataSourceSet`
+- The `Engine` itself
+
+## Plugin Set
+
+The `PluginSet` is a collection of all of your plugins and extensions that can
+be used to process data from data sources (i.e. files.) You can create a new
```
mslukebo, line 16: "i.e." -> "e.g."

documentation/Using-the-SDK/Using-the-engine.md
```diff
@@ -0,0 +1,104 @@
...
+
+Finally, you can specify that the `DataSourceSet` should take ownership of the
+`PluginSet`. This means that when the `DataSourceSet` is disposed, the
+`DataSourceSet` will also dispose the `PluginSet`. If you wish the `PluginSet`
+to be used many times, then use the `DataSourceSet.Create(PluginSet, bool)`
+overload. By default, the `DataSourceSet` will take ownership of the `PluginSet`
+so you *MUST* use the overload if you do not want this to occur.
+
+NOTE: at this time, reusing the `PluginSet` is not fully implemented, and so
+`Create(PluginSet, false)` will throw a `NotSupportedException`. The ability
```
mslukebo, line 63: should that be a NotImplementedException instead?

documentation/Using-the-SDK/Using-the-engine.md
```diff
@@ -0,0 +1,104 @@
...
+instance in order to create a useable `Engine.` Similar to how the `DataSourceSet`
+can take ownership of the `PluginSet`, the `Engine` takes ownership of the
+`DataSourceSet`. Set the `OwnsDataSources` property on the `EngineCreateInfo`
+instance to suppress this behavior.
+Once your `Engine` has been created, you can enable cookers and tables to
+participate in processing. If an attempt is made to enable a cooker for which
+there is no corresponding data source in the `DataSourceSet`, then an exception
+will be thrown. You may use the Try versions of the Enable methods to avoid
+the exceptions.
+Once you are read, call process and then examine the results.
```
mslukebo, line 90: "Once you are read" ?

documentation/Using-the-SDK/Using-the-engine.md
```diff
@@ -0,0 +1,104 @@
...
+
+````cs
+using var plugins = PluginSet.Load();
+using var dataSources = DataSourceSet.Create(plugins);
```
mslukebo, line 94: might be nice to have a ctor that takes in a `params IDataSource[]`

documentation/Using-the-SDK/Using-the-engine.md
```diff
@@ -0,0 +1,104 @@
...
+dataSources.Add(new FileDataSource("myfile.txt");
+
+var createInfo = new EngineCreateInfo(dataSources);
```
mslukebo, line 97: it seems odd to me to have a class named "Info" that's really also the factory/main entry point. To me, an "Info" class is an immutable collection of information, and nothing else. 

Maybe we should rename this class?

documentation/Using-the-SDK/Using-the-engine.md
```diff
@@ -0,0 +1,104 @@
...
+using var engine = Engine.Create(info);
+
+engine.EnableCooker(DataCookerPath);
```
mslukebo, line 100: Why is this method on `Engine` instead of `EngineCreationInfo`?

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Repository/DataExtensionRepositoryExtensions.cs
```diff
@@ -17,6 +17,17 @@ namespace Microsoft.Performance.SDK.Runtime.Extensibility.DataExtensions.Reposit
     /// </summary>
     public static class DataExtensionRepositoryExtensions
     {
+        // todo: use a new class instead of n out parameters
```
mslukebo, line 20: did you mean to leave this here?

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Tables/TableExtensionSelector.cs
```diff
@@ -19,11 +19,14 @@ public class TableExtensionSelector
...
 
         /// <summary>
-        ///  Constructor
+        ///  Creates a new instance of the <see cref="TableExtensionSelector" /> class.
         /// </summary>
         /// <param name="dataExtensions">
         ///     Repository of data extensions.
         /// </param>
+        /// <exception cref="ArgumentNullException">
+        ///     <paramref name="self"/> is <c>null</c>.
```
mslukebo, line 28: what is self?

src/Microsoft.Performance.Toolkit.Engine/DataSourceSetExtensions.cs
```diff
@@ -11,9 +11,9 @@ namespace Microsoft.Performance.Toolkit.Engine
 {
     /// <summary>
     ///     Contains static (Shared in Visual Basic) methods for interacting
```
mslukebo, line 13: can we get rid of "(Shared in Visual Basic)"

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -19,108 +15,31 @@ public sealed class EngineCreateInfo
...
         ///     Initializes a new instance of the <see cref="EngineCreateInfo"/> class.
         /// </summary>
-        /// <param name="extensionDirectories">
-        ///     The extension directories from which the runtime instance is to load plugins.
+        /// <param name="dataSources">
+        ///     The data sources to be processed in the engine.
         /// </param>
         /// <exception cref="ArgumentNullException">
-        ///     Thrown is <paramref name="extensionDirectories"/> is <c>null</c>.
+        ///     Thrown is <paramref name="dataSources"/> is <c>null</c>.
```
mslukebo, line 34: thrown if*

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -141,5 +60,23 @@ public EngineCreateInfo(IEnumerable<string> extensionDirectories)
...
+
+        /// <summary>
+        ///     Gets the data sources that are to be processed by an <see cref="Engine"/>
+        ///     instance.
+        /// </summary>
+        public DataSourceSet DataSources { get; }
+
+        /// <summary>
+        ///     Gets or sets a value indicating whether the created <see cref="Engine"/>
+        ///     instance owns the data sources passed in <see cref="DataSources"/>.
```
mslukebo, line 72: I would either say "passed in (to the ctor)" or just "owns <see cref="DataSources"/>

src/Microsoft.Performance.Toolkit.Engine/PluginSet.cs
```diff
@@ -0,0 +1,375 @@
...
+        ///     loader will be used.
+        ///     <remarks>
+        ///         The default loader provides no isolation.
+        ///     </remarks>
+        /// </param>
+        /// <returns>
+        ///     A new instance of the <see cref="PluginSet"/> class containing all
+        ///     of the successfully discovered plugins. The returned instance will
+        ///     also contain a collection of non-fatal errors that occurred when
+        ///     creating this data set (i.e. a plugin failed to load.)
```
mslukebo, line 241: e.g.

src/Microsoft.Performance.Toolkit.Engine/PluginSet.cs
```diff
@@ -0,0 +1,375 @@
...
+        /// </exception>
+        /// <exception cref="InvalidExtensionDirectoryException">
+        ///     One or more directory paths in <paramref name="extensionDirectories"/>
+        ///     is invalid or does not exist.
+        /// </exception>
+        public static PluginSet Load(
+            IEnumerable<string> extensionDirectories,
+            IAssemblyLoader assemblyLoader)
+        {
+            Guard.NotNull(extensionDirectories, nameof(extensionDirectories));
```
mslukebo, line 257: Is there a reason you didn't want to use the PluginsLoader? We can isolate components more if we just make this PluginSet a consumer that passes directories down to the loader.

documentation/Using-the-SDK/Using-the-engine.md
zachnew-msft: Should be ready, thanks!

documentation/Using-the-SDK/Using-the-engine.md
zachnew-msft: I choose not supported as we aren't supporting it yet, but the method is in general implemented.

documentation/Using-the-SDK/Using-the-engine.md
zachnew-msft: Thanks!

documentation/Migration/Update-0.109.md
zachnew-msft: For newer versions of C#, they don't need them. However, for consistency with other docs, I will put them in. Thanks!

documentation/Using-the-SDK/Using-the-engine.md
zachnew-msft: It is loosely modeled after ProcessStartInfo where you set everything up first. This class contains all of the information needed to create an Engine instance. It contains additional parameters that users may or may not want to set. There are direct factory methods on the Engine for more streamlined use cases.

src/Microsoft.Performance.Toolkit.Engine/DataSourceSetExtensions.cs
zachnew-msft: I wrote these comments following those placed on extension method classes in the framework.

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
zachnew-msft: Thanks!

src/Microsoft.Performance.Toolkit.Engine/PluginSet.cs
zachnew-msft: I just moved the code out of the engine as it was. I can look at doing that.

documentation/Using-the-SDK/Using-the-engine.md
zachnew-msft: Because we don't force users to use EngineCreateInfo unless they really need to configure many things. I will add examples that use the simpler factory methods.

documentation/Using-the-SDK/Using-the-engine.md
mslukebo: Ah I misread this as `using var engine = EngineCreateInfo.Create()` last time. Nevermind!

documentation/Using-the-SDK/Using-the-engine.md
```diff
@@ -0,0 +1,104 @@
...
+
+// ... and so on
+
+````
+
+If an attempt is made to add an `IDataSource` to the `DataSourceSet` for which
+no plugin can process, then the `DataSourceSet` will throw an exception. The
+`DataSourceSet` class provides Try versions of each of the methods (e.g.
+`TryAddDataSource`) that will not throw if the `IDataSource` does not have a
+corresponding plugin that can process the data source.
```
anramanath, line 53: ...will not throw if the `IDataSource` does not have a plugin to process the data source.

is a little simpler. I got confused with the negations

src/Microsoft.Performance.Toolkit.Engine.Tests/DataSourceSetTests.cs
```diff
@@ -0,0 +1,581 @@
...
+using Microsoft.Performance.Toolkit.Engine.Tests.TestCookers.Source4;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Performance.Toolkit.Engine.Tests
+{
+    [TestClass]
+    public class DataSourceSetTests
+        : EngineFixture
+    {
+        private DataSourceSet Sut { get; set; }
```
anramanath, line 19: What does `Sut` mean? Can we rename this to something more descriptive?

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -312,271 +165,257 @@ public IEnumerable<TableDescriptor> AvailableTables
...
         /// <exception cref="ArgumentNullException">
-        ///     <paramref name="dataSources"/> is <c>null</c>.
-        ///     - or -
-        ///     <paramref name="processingSourceType"/> is <c>null</c>.
+        ///     <paramref name="createInfo"/> is <c>null</c>.
         /// </exception>
-        /// <exception cref="InstanceAlreadyProcessedException">
-        ///     This instance has already been processed.
+        /// <exception cref="EngineException">
+        ///     An fatal error occurred while creating the engine.
```
anramanath, line 360: A fatal error*

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -312,271 +165,257 @@ public IEnumerable<TableDescriptor> AvailableTables
...
-        /// <param name="processingSourceType">
-        ///     The <see cref="IProcessingSource"/> to use to process <paramref name="dataSource"/>.
-        /// </param>
-        /// <returns>
-        ///     <c>true</c> if the <see cref="IDataSource"/> has been added for processing by the <see cref="IProcessingSource"/>;
-        ///     <c>false</c> otherwise. Note that <c>false</c>
-        ///     is always returned when <see cref="IsProcessed"/> is <c>true</c>.
-        /// </returns>
+        /// <exception cref="EngineException">
+        ///     An fatal error occurred while creating the engine.
```
anramanath, line 313: A fatal error occurred*

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -647,28 +502,70 @@ public bool TryEnableCooker(DataCookerPath dataCookerPath)
...
+                        false,
+                        "If allTables contains the descriptor, then it should be found or there's a bug.");
+                }
+
+                if (tableReference.Availability != DataExtensionAvailability.Available)
+                {
+                    throw new TableException($"The requested table is not available: {tableReference.Availability}.");
+                }
+
+                // todo: we need to check the "filtered repository" to make sure that it's available
```
anramanath, line 550: is this todo still valid?

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -898,133 +773,116 @@ protected ILogger CreateLogger(Type type)
                     }
                     catch (Exception e)
                     {
-                        //
-                        // todo: log
-                        //
-
-                        Console.Error.WriteLine(e);
+                        logger.Error($"Failed to initialize processing source '{cds.Name}': {e}");
```
anramanath, line 776: thanks for changing to logger.

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Repository/DataExtensionRepositoryExtensions.cs
zachnew-msft: No I didn't, I can remove it.

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Tables/TableExtensionSelector.cs
zachnew-msft: I will document it.

documentation/Using-the-SDK/Using-the-engine.md
zachnew-msft: Thanks, I tend to get the rules for parentheses and quotation marks confused.

src/Microsoft.Performance.Toolkit.Engine.Tests/DataSourceSetTests.cs
zachnew-msft: Sut means "System Under Test." It is the thing that is being tested.

documentation/Using-the-SDK/Using-the-engine.md
zachnew-msft: That leads to interesting edge cases. For example, what if one of the passed data sources isn't supported by any of the plugins. Do we throw an exception? If so, does the half created data source set then dispose the plugins that were passed in if it took ownership? We can always add that method in the future.

src/Microsoft.Performance.Toolkit.Engine/DataSourceSet.cs
```diff
@@ -0,0 +1,587 @@
...
+            this.isDisposed = false;
+        }
+
+        /// <summary>
+        ///     Gets the collection of Data Sources that are to be processed by a specific <see cref="IProcessingSource"/>.
+        /// </summary>
+        /// <exception cref="ObjectDisposedException">
+        ///     This instance is disposed.
+        /// </exception>
+        public IReadOnlyDictionary<ProcessingSourceReference, IReadOnlyList<IReadOnlyList<IDataSource>>> DataSourcesToProcess
```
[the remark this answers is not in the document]
zachnew-msft: Yes, that makes sense.

src/Microsoft.Performance.Toolkit.Engine/DataSourceSet.cs
```diff
@@ -0,0 +1,587 @@
...
+        ///     Seals this instance, preventing any new data sources from being added.
+        /// </summary>
+        /// <returns>
+        ///    A reference to this instance after the operation has completed.
+        /// </returns>
+        /// <exception cref="ObjectDisposedException">
+        ///     This instance is disposed.
+        /// </exception>
+        public DataSourceSet Seal()
+        {
```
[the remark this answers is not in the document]
zachnew-msft: Yes, thanks. Updated.

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
zachnew-msft: I'll double check

src/Microsoft.Performance.Toolkit.Engine/PluginSet.cs
```diff
@@ -0,0 +1,375 @@
...
+            }
+        }
+
+        /// <summary>
+        ///     Gets all of the loaded Processing Sources.
+        /// </summary>
+        /// <exception cref="ObjectDisposedException">
+        ///     This instance is disposed.
+        /// </exception>
+        public IEnumerable<ProcessingSourceReference> ProcessingSourceReferences => this.Extensions.PlugIns;
```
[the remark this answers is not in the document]
zachnew-msft: These are public so that users can examine what has been loaded.

src/Microsoft.Performance.Toolkit.Engine/PluginSet.cs
zachnew-msft: It is, thanks

src/Microsoft.Performance.Toolkit.Engine/PluginSet.cs
zachnew-msft: The pluginsloader is in an assembly that uses .NET Core, and the Engine is .NET standard. The Engine is .NET standard to facilitate use from many different places / scenarios, and there are no plans to upgrade it in the immediate / near future.

src/Microsoft.Performance.Toolkit.Engine/DataSourceSet.cs
```diff
@@ -0,0 +1,587 @@
...
+        ///     directory.
+        /// </summary>
+        public static DataSourceSet Create()
+        {
+            PluginSet plugins = null;
+            try
+            {
+                plugins = PluginSet.Load();
+                return Create(plugins);
+            }
```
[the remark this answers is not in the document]
jmaxson-ms: Ok, I can see that. If ownsPlugin is true, but we throw, then we don't cleanup the PluginSet? That's a weird scenario.

src/Microsoft.Performance.Toolkit.Engine/DataSourceSet.cs
```diff
@@ -0,0 +1,537 @@
...
+        ///     The Data Sources to process.
+        /// </param>
+        /// <param name="processingSourceType">
+        ///     The <see cref="IProcessingSource"/> to use to process the <paramref name="dataSources"/>.
+        /// </param>
+        /// <returns>
+        ///    A reference to this instance after the operation has completed.
+        /// </returns>
+        /// <exception cref="ArgumentNullException">
+        ///     <paramref name="dataSources"/> is <c>null</c>.
```
jmaxson-ms, line 327: ><paramref name="dataSources"/> is <c>null</c>. [](http://example.com/codeflow?start=16&length=46)

or any element of dataSources is null

src/Microsoft.Performance.Toolkit.Engine/DataSourceSet.cs
```diff
@@ -0,0 +1,537 @@
...
+                return false;
+            }
+        }
+
+        /// <summary>
+        ///     Seals this instance, preventing any new data sources from being added.
+        /// </summary>
+        /// <returns>
+        ///    A reference to this instance after the operation has completed.
+        /// </returns>
```
jmaxson-ms, line 403: This comment needs to be updated.

src/Microsoft.Performance.Toolkit.Engine/DataSourceSet.cs
```diff
@@ -0,0 +1,537 @@
...
+        {
+            Debug.Assert(!this.isDisposed);
+
+            Debug.Assert(dataSources != null);
+            Debug.Assert(processingSourceType != null);
+            Debug.Assert(processingSourceReferences != null);
+            Debug.Assert(dataSourcesToProcess != null);
+            Debug.Assert(typeIs != null);
+
+            var cdsr = processingSourceReferences.FirstOrDefault(x => typeIs(x.Instance.GetType(), processingSourceType));
```
jmaxson-ms, line 479: >FirstOrDefault(x => typeIs(x.Instance.GetType(), processingSourceType)); [](http://example.com/codeflow?start=50&length=72)

shouldn't we check all processing sources that might support the given source?

jmaxson-ms: nevermind, I read this wrong

src/Microsoft.Performance.Toolkit.Engine/DataSourceSet.cs
```diff
@@ -0,0 +1,537 @@
...
+                return first.Is(second);
+            }
+        }
+
+        private void AddDataSourcesCore(
+            IEnumerable<IDataSource> dataSources,
+            Type processingSourceType,
+            List<ProcessingSourceReference> processingSourceReferences,
+            Dictionary<ProcessingSourceReference, List<List<IDataSource>>> dataSourcesToProcess,
+            Func<Type, Type, bool> typeIs)
```
jmaxson-ms, line 469: >typeIs [](http://example.com/codeflow?start=35&length=6)

why is 'typeIs' being passed in as a function?

jmaxson-ms: was this method originally static or something? I think several of these parameters could be removed and accessed directly from 'this'

jmaxson-ms: Weird that not all comments from the other PR tool seem to be added here. so... I'll copy/paste:

was this method originally static or something? I think several of these parameters could be removed and accessed directly from 'this'

src/Microsoft.Performance.Toolkit.Engine/EngineException.cs
```diff
@@ -628,4 +628,98 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
...
+    ///     Represents the error that occurs when an attempt is made to enable a
+    ///     source cooker on the engine when the data sources do not provide the
+    ///     data required for said cooker.
+    ///     <para />
+    ///     An example of this condition would be when you attempt to add a cooker
+    ///     that cooks events from an ETL file (Event Tracing for Windows trace file),
+    ///     but no ETL files are present in the set of files to process.
+    /// </summary>
+    [Serializable]
+    public class NoInputDataException
```
jmaxson-ms, line 642: >NoInputDataException [](http://example.com/codeflow?start=17&length=20)

Should this be NoDataSourceException?

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -55,73 +38,42 @@ public abstract class Engine
...
-        private IReadOnlyCollection<string> extensionDirectories;
         private bool isProcessed;
-        private IEnumerable<ErrorInfo> creationErrors;
 
         private IApplicationEnvironment applicationEnvironment;
-        private IAssemblyLoader loader;
         private bool isDisposed;
 
-        internal Engine(Func<Type, ILogger> loggerFactory)
+        internal Engine(EngineCreateInfo createInfo, DataSourceSet internalDataSourceSet)
```
jmaxson-ms, line 55: >internalDataSourceSet [](http://example.com/codeflow?start=67&length=21)

would you mind adding comment about internalDataSourceSet

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -647,28 +502,68 @@ public bool TryEnableCooker(DataCookerPath dataCookerPath)
...
         /// <exception cref="ObjectDisposedException">
         ///     This instance is disposed.
         /// </exception>
+        /// <exception cref="TableException">
+        ///     A table is not available <paramref name="tableDescriptor"/>.
+        /// </exception>
+        /// <exception cref="TableNotFoundException">
+        ///     A table cannot be found for the given <paramref name="tableDescriptor"/>.
+        /// </exception>
         public void EnableTable(TableDescriptor descriptor)
```
jmaxson-ms, line 520: Looks like there's a possible ArgumentException too

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -1266,35 +1127,41 @@ private sealed class RuntimeMessageBox
             {
                 var message = string.Format(formatProvider, format, args);
 
-                Console.Error.WriteLine(
+                // todo: use the logger here too
+
+                this.logger.Error(
                     string.Concat("[", icon.ToString(), "]: ", message));
```
jmaxson-ms, line 1133: should this just call into the other Show method?

src/Microsoft.Performance.Toolkit.Engine/DataSourceSet.cs
zachnew-msft: Yes it was static. I will look at using this.

src/Microsoft.Performance.Toolkit.Engine/EngineException.cs
zachnew-msft: I was trying to communicate that the source cooker did not have any input. I can rename, I think that makes sense.

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
zachnew-msft: Sure

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
zachnew-msft: Thanks!

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
zachnew-msft: We could, but what buttons and caption would be passed? I think this is best addressed as part of issue #109 .

jmaxson-ms (approved):

src/Microsoft.Performance.Toolkit.Engine.Tests/DataSourceSetTests.cs
anramanath: cool. I didn't know that. Thanks.

anramanath (approved):

documentation/Using-the-SDK/Using-the-engine.md
```diff
@@ -0,0 +1,110 @@
...
+to be used many times, then use the `DataSourceSet.Create(PluginSet, bool)`
+overload. By default, the `DataSourceSet` will take ownership of the `PluginSet`
+so you *MUST* use the overload if you do not want this to occur.
+
+NOTE: at this time, reusing the `PluginSet` is not fully implemented, and so
+`Create(PluginSet, false)` will throw a `NotSupportedException`. The ability
+to pass `false` will be added in a future update.
+
+````cs
+using (var plugins1 = PluginSet.Load())
```
mslukebo, line 70: this might be a dumb question, but why even use a `using` here? If the plugins set is disposed after the data source set is disposed, the using is a red herring. It implies "everything in this block has this thing available to use, and it will be disposed after it."

documentation/Using-the-SDK/Using-the-engine.md
```diff
@@ -0,0 +1,110 @@
...
+using (var dataSources = DataSourceSet.Create(plugins))
+{
+    dataSources.Add(new FileDataSource("myfile.txt");
+
+    var createInfo = new EngineCreateInfo(dataSources);
+    using (var engine = Engine.Create(info))
+    {
+        engine.EnableCooker(DataCookerPath);
+
+        var results = engine.Process();
```
mslukebo, line 106: suggestion: this code makes it seems the Engine (and data source set and plugin set) need to still be around to access the result. I'm assuming this is not true; e.g. we can use the result even after the engine is disposed?

If that's true, maybe declare the results _before_ the `using`s so it's clear that's the case

documentation/Using-the-SDK/Using-the-engine.md
zachnew-msft: This is the same pattern as a Stream and a StreamReader. These are well established patterns. You pass false to DataSourceSet.Create to not dispose plugin set.

documentation/Using-the-SDK/Using-the-engine.md
zachnew-msft: For now that is the case due to the internals of ````result````. We can look at removing that restriction in the future as it would not be a breaking change.

mslukebo (dismissed):

jmaxson-ms (approved):

anramanath (approved):

PR #128, opened 2021-09-08
Remove defunct code for RC1.

jmaxson-ms:
This removes obsolete code related to DataCookerPaths and DataOutputPaths.


zachnew-msft (approved):

zachnew-msft (approved):

documentation/Migration/Update-1.0.0-rc.md
```diff
@@ -97,6 +97,57 @@ catch (UnsupportedDataSourceException)
 }
```
mslukebo, line 97: update to target rc1.md file :)

src/Microsoft.Performance.SDK/Extensibility/DataCookerPath.cs
```diff
@@ -12,21 +12,13 @@ namespace Microsoft.Performance.SDK.Extensibility
...
-        /// </summary>
-        [Obsolete("Fully qualified data cooker paths will no longer be supported in SDK v1.0.0 release candidate 1")]
-        public static string Format => "SourceId/CookerId";
-
         /// <summary>
         ///     If the data cooker is not a source data cooker, this is the source parser Id.
         /// </summary>
-        [Obsolete("This will be removed by SDK v1.0.0 release candidate 1")]
-        public const string EmptySourceParserId = "";
+        internal const string EmptySourceParserId = "";
```
mslukebo, line 18: why not `= DataCookerPathInternal.EmptySourceParserId` or just use that directly

src/Microsoft.Performance.SDK/Processing/RequiresCookerAttribute.cs
```diff
@@ -12,83 +11,14 @@ namespace Microsoft.Performance.SDK.Processing
     /// or on individual <see cref="TableDescriptor"/> static properties to indicate that a 
     /// table described by the <see cref="TableDescriptor"/> requires a data cooker.
     /// </summary>
-    [Obsolete("This will be removed by SDK v1.0.0 release candidate 1. Please use RequiresSourceCooker or RequiresCompositeCooker attributes instead.")]
-    public class RequiresCookerAttribute
+    public abstract class RequiresCookerAttribute
```
mslukebo, line 14: should this be internal? it was public before since it was used by plugins directly

mslukebo (changes requested):
Don't we also need to remove obsolete stuff from `CustomDataSource*` classes and `TableAttribute` ?

src/Microsoft.Performance.SDK/Extensibility/DataCookerPath.cs
jmaxson-ms: If it's internal, then we'd have to make internals visible to the runtime. Not something I want to do to solve this "problem".

mslukebo: Makes sense. Should we add comments above the two definitions then saying this value must match the other? Just worried one day we'll decide to update the empty source parser ID and get some weird bugs.

PR #129, opened 2021-09-09
Composite cookers are no longer stored globally.

jmaxson-ms:


src/Microsoft.Performance.SDK.Runtime/Extensibility/CustomDataProcessorExtensibilitySupport.cs
```diff
@@ -37,35 +39,86 @@ public class CustomDataProcessorExtensibilitySupport
...
                 {
-                    this.tableReferences[tableDescriptor] = tableReference;
-                    return true;
+                    throw new ExtensionTableException(
+                        $"The internal table {tableReference.TableDescriptor.Type} references source parser " +
+                        $"{cookerPath.SourceParserId} and cannot be enabled on {this.dataProcessor.SourceParserId}.");
                 }
             }
 
+            this.tableReferences[tableDescriptor] = tableReference;
```
zachnew-msft, line 107: I noticed we have a lock earlier in the methods, which suggests that this may be called from multiple threads. Should we be protecting this update here?

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Tables/ITableExtensionReference.cs
```diff
@@ -15,5 +15,9 @@ public interface ITableExtensionReference
           IDataExtensionDependency,
           IDisposable
     {
+        /// <summary>
+        ///     Gets a value indicating whether this table is an internal table.
```
zachnew-msft, line 19: Have we defined what internal table means anywhere in the docs?

src/Microsoft.Performance.Toolkit.Engine.Tests/ToolkitEngineTests.cs
```diff
@@ -534,6 +534,16 @@ public void BuildTable_SingleRowTable()
             }
         }
 
+        //[TestMethod]
+        //[FunctionalTest]
```
zachnew-msft, line 538: Do we still need this?

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Tables/ITableExtensionReference.cs
jmaxson-ms: I guess it depends on what docs you're asking about.

On TableAttribute:
        /// <param name="internalTable">
        ///     When this is set, the table will not be exposed through the data extension repository.
        ///     Note that the <see cref="IProcessingSource"/> that owns this will be required to create this table
        ///     when this is set.
        /// </param>

src/Microsoft.Performance.Toolkit.Engine.Tests/ToolkitEngineTests.cs
jmaxson-ms: Thanks, I'll take a look. The problem with working on this on/off for 1.5 months...

jmaxson-ms: A couple tests for internal tables have been added now.

documentation/Migration/Update-1.0.0-rc.md
```diff
@@ -97,6 +97,36 @@ catch (UnsupportedDataSourceException)
...
+
+- The `AddTable` method becomes `EnableTable` or `TryEnableTable`.
+
+- The `GetAllRequiredSourceDataCookers` was renamed to `GetRequiredSourceDataCookers`.
+
+- The `GetAllRequiredTables was renamed to `GetEnabledInternalTables`.
+
+### IDataExtensionRetrievalFactory
+
+This was moved from the SDK to SDK.Runtime.
```
mslukebo, line 118: nit: "SDK namespace" to the "SDK.Runtime" namespace

documentation/Migration/Update-1.0.0-rc.md
```diff
@@ -97,6 +97,36 @@ catch (UnsupportedDataSourceException)
...
+
+## TableAttribute and InternalTableAttribute
+
+- The `InternalTableAttribute` was removed.
+
+- Internal table constructor arguments and properties have been removed from `TableAttribute`.
```
mslukebo, line 124: Doesn't https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/117/files do this? Can we just get that PR in to reduce the number of changes this PR is making?

src/Microsoft.Performance.SDK.Runtime.Tests/Extensibility/CompositeCookerRepositoryTests.cs
```diff
@@ -0,0 +1,93 @@
...
+                sourceCookerData,
+                this.Sut,
+                this.dataExtensionRepository);
+
+            this.Sut.Initialize(retrievalFactory);
+
+            var cookerOutput = this.Sut.GetCookerOutput(dataCookerReference.Path);
+            Assert.AreEqual((ICookedDataRetrieval)testCompositeCooker, cookerOutput);
+
+            // if this is called a second time, it will throw an exception
```
mslukebo, line 85: ideally this would be a separate test method

src/Microsoft.Performance.SDK.Runtime/Extensibility/CustomDataProcessorExtensibilitySupport.cs
```diff
@@ -37,35 +40,94 @@ public class CustomDataProcessorExtensibilitySupport
...
+            }
+
+            if (uniqueParserIds.Any())
+            {
+                var requiredParserId = uniqueParserIds.First();
+                if (!DataCookerPath.EqualityComparer.Equals(requiredParserId, this.dataProcessor.SourceParserId))
                 {
-                    this.tableReferences[tableDescriptor] = tableReference;
-                    return true;
+                    throw new WrongProcessorTableException(
```
mslukebo, line 110: mind adding a comment explaining what's being checked here?

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/ProcessingSystemCompositeCookers.cs
```diff
@@ -0,0 +1,240 @@
...
+
+        private ICookedDataRetrieval GetOrCreateCompositeCooker(
+            DataCookerPath cookerPath,
+            Func<DataCookerPath, IDataExtensionRetrieval> createDataRetrieval)
+        {
+            Debug.Assert(createDataRetrieval != null);
+
+            this.ThrowIfDisposed();
+
+            Debug.Assert(
```
mslukebo, line 90: why not just throw an exception

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/ProcessingSystemCompositeCookers.cs
```diff
@@ -0,0 +1,240 @@
...
+
+                this.dataExtensionRepository = null;
+                this.retrievalFactory = null;
+                this.disposedValue = true;
+            }
+        }
+
+        /// <summary>
+        ///     This class uses <see cref="ProcessingSystemCompositeCookers"/> to create and store composite
+        ///     cookers, but uses a its own <see cref="IDataExtensionRetrievalFactory"/> which has access
```
mslukebo, line 171: uses its own factory for what?

src/Microsoft.Performance.SDK.Tests/DataTypes/TestRecord.cs
```diff
@@ -14,12 +14,12 @@ public class TestRecord
 
         public int CompareTo(int otherKey)
         {
-            return this.Key.CompareTo(otherKey);
+            return Key.CompareTo(otherKey);
         }
 
         public int GetKey()
         {
-            return this.Key;
```
mslukebo, line 22: why this change? we usually keep this prefix on properties I thought

src/Microsoft.Performance.SDK/Processing/TableDescriptor.cs
```diff
@@ -139,8 +144,24 @@ public sealed class TableDescriptor
...
+        /// <remarks>
+        ///     Metadata tables are automatically marked as internal (<see cref="IsInternalTable"/>).
+        /// </remarks>
         public bool IsMetadataTable { get; }
 
+        /// <summary>
+        ///     Gets a value indicating whether this instance describes an internal table.
+        /// </summary>
+        /// <remarks>
+        ///     Internal tables are not required to have a static build method.
```
mslukebo, line 156: I don't understand what an internal table is then. Why does it not need a static build method?

documentation/Migration/Update-1.0.0-rc.md
jmaxson-ms: Actually, this change is a bit different. This removes internal from the table attribute all together. Rather than putting it on the table attribute, it is stored in the table descriptor. The table descriptor is what is stored longterm and should fully describe the table. I'm not sure how the table attributes and table descriptors evolved over time, but this is the correct place for it today.

src/Microsoft.Performance.SDK/Processing/TableDescriptor.cs
jmaxson-ms: An internal table is entirely maintained by the processing source and data processor. They are responsible for initializing and building the table, and therefore they can build it however they please. An internal table can still take advantage of the benefits that an extension table provides - making sure that a table's dependencies are all available, and restricting the table to data from those dependencies.

I'm not sure that we need to keep the internal table feature, but I also want to be careful about removing it as a feature so close to SDK 1.0. And even if we did that, I'd want to do it in a separate PR.

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/ProcessingSystemCompositeCookers.cs
jmaxson-ms: this is a private method, and if we're hitting this, then it's a bug in our code.

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/ProcessingSystemCompositeCookers.cs
```diff
@@ -0,0 +1,240 @@
...
+                Debug.Assert(false);
+                throw new InvalidOperationException(
+                    $"Failed to create {nameof(IDataExtensionRetrieval)} needed for the composite cooker.");
+            }
+
+            var compositeCooker = compositeCookerReference.CreateInstance(compositeCookerRequiredData);
+            if (compositeCooker == null)
+            {
+                Debug.Assert(false, "If the data cooker reference is available, this shouldn't fail.");
+                throw new InvalidOperationException(
```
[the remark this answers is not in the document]
jmaxson-ms: i'm going to remove this exception and leave the assert. this should never happen.

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/ProcessingSystemCompositeCookers.cs
jmaxson-ms: I think the comments on IDataExtensionRetrievalFactory should answer this question. We can chat if you disagree.

src/Microsoft.Performance.SDK.Runtime/Extensibility/FilteredDataRetrieval.cs
```diff
@@ -18,17 +18,47 @@ namespace Microsoft.Performance.SDK.Runtime.Extensibility
...
+        ///     Creates an instance of the <see cref="FilteredDataRetrieval"/> class for a given data extension.
+        /// </summary>
+        /// <param name="sourceCookerData">
+        ///     Provides access to source cooker data.
+        /// </param>
+        /// <param name="compositeCookers">
+        ///     Provides access to composite cookers.
+        /// </param>
+        /// <param name="dataExtensionRepository">
+        ///     Provides access to data processors (to be removed for V1).
```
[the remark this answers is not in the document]
jmaxson-ms: I'll add the issue #

src/Microsoft.Performance.SDK.Tests/DataTypes/TestRecord.cs
jmaxson-ms: not sure why this is here, I'll revert. thanks.

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -17,12 +17,17 @@
 using Microsoft.Performance.SDK.Runtime.Extensibility.DataExtensions.Repository;
 using Microsoft.Performance.SDK.Runtime.Extensibility.DataExtensions.Tables;
 
+// TODO::support concurrency in the Engine
```
anramanath, line 20: I noticed a lot of todos on this PR. Do we want these things for 1.0 or later? Also maybe we can tag/create the todos with the work item?

PR #134, opened 2021-09-14
RC1 -> main

mslukebo:
This is a separate branch because there's merge conflicts and GitHub wouldn't let me resolve them on the web interface


documentation/Migration/Overview.md
```diff
@@ -3,5 +3,4 @@
 These documents outline how to update your code due to breaking changes between major versions.
 
 - [Updating to 0.109.*](./Update-0.109.md)
-- [Updating to 1.0.0-rc](./Update-1.0.0-rc.md)
```
jmaxson-ms, line 6: There is already documentation in here from a previous check-in. That will need to be moved to the new rc1.md file.

zachnew-msft (dismissed):

documentation/Migration/Overview.md
mslukebo: Good catch. Merged the files

zachnew-msft (approved):

jmaxson-ms (approved):

Issue #33, opened 2021-01-25
[Feature Request] Support a way for a CDS to whitelist tables


mslukebo:
Currently, all tables defined in a given assembly is attempted to be built (either by passing it into a Custom Data Processor's `BuildTableCore` or getting data from a data cooker and explicitly calling a static `BuildTable` method).

We should provide a way for a CDS to explicitly list the tables outputted by a plugin. By default, all tables defined in the CDS' assembly can be assumed to be this output. If this list is included on the CDS, only those tables enumerated are considered part of this CDS' output. If a table is not outputted by any CDS in its assembly, it should never attempt to be built.

The reason this would make sense is because tables are fundamentally part of a CDS' interface: data sources in, tables out. It does not make sense for a table to show up if it's not part of any CDS contract.

Some possible ways to do this are:
1) Provide a `TablesOutputAttribute` that a CDS can decorate itself with. The attribute would take a `params` of `Type`s that reference the tables the CDS is whitelisting
2) Allow CDSs to expose a `static IEnumerable<TableDescriptor> TablesOutput` property that references the tables the CDS is whitelisting
3) Allow CDSs to override a `EnableTables()` method that by default enables all tables in its assembly

I personally like option 1, since the entire CDS contract (data sources in, tables out) can be declared with attributes.

trgibeau:
Follow up

mslukebo:
- Introduce a new `ITableDiscoverer`
- Include implementations such as `AllTablesInAssemblyTableDiscoverer` and `AllTablesInNamespaceTableDiscoverer`
- `ProcessingSource` will have a new ctor that takes in an `ITableDiscoverer`
- old `ProcessingSource` ctor will defer to using an `AllTablesInAssemblyTableDiscoverer`
- remove/obsolete the ctor that takes a `Func` for additional tables

mslukebo:
- Introduce a new `ITableDiscoverer`
- Include implementations such as `AllTablesInAssemblyTableDiscoverer` and `AllTablesInNamespaceTableDiscoverer`
- `ProcessingSource` will have a new ctor that takes in an `ITableDiscoverer`
- old `ProcesisngSource` ctor will defer to using a

PR #136, opened 2021-10-01
User/zachnew msft/table discovery

zachnew-msft:
Implements #33

### DIFF
--- a/documentation/Migration/Update-1.0.0-rc.md
+++ b/documentation/Migration/Update-1.0.0-rc.md
@@ -9,6 +9,23 @@ into two sections: [Breaking Changes](#breaking-changes) and
 
 There are a number of breaking changes in this version; please see the release notes for a list of these changes.
 
+## SDK
+
+The following are required if you are using the `ProcessingSource` base class and
+are using the cosntructors that take `additionalTablesProvider` and/or the
+`tableAssemblyProvider` parameters.
+
+These parameters are being removed and replaced by the `ITableProvider` interface.
+If you have custom logic for determining the tables exposed by a `ProcessingSource`,
+you must implement the new interface.
+
+The default behavior of using all tables found in the assembly has been preserved.
+This change only effects those `ProcessingSource`s that use custom table providers.
+
+Two helper methods have been added for use:
+`TableDiscovery.CreateForAssembly` and `TableDiscovery.CreateForNamespace.` User's
+may also provide their own implementations.
+
 ## Engine
 
 The following are required if you are using the `Engine`:
--- a/src/Microsoft.Performance.SDK.Tests/TableDiscoveryTests.cs
+++ b/src/Microsoft.Performance.SDK.Tests/TableDiscoveryTests.cs
@@ -1,0 +1,170 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Linq;
+using Microsoft.Performance.SDK.Processing;
+using Microsoft.Performance.Testing;
+using Microsoft.Performance.Testing.SDK;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Performance.SDK.Tests
+{
+    public class TableDiscoveryTests
+    {
+        [TestMethod]
+        [UnitTest]
+        public void NoTablesInAssemblyLeavesEmptyProperties()
+        {
+            var assembly = new FakeAssembly
+            {
+                TypesToReturn = new[]
+                {
+                    typeof(DateTime),
+                }
+            };
+
+            var sut = TableDiscovery.CreateForAssembly(assembly);
+            var tables = sut.Discover(new FakeSerializer());
+
+            Assert.AreEqual(0, tables.Count);
+        }
+
+        [TestMethod]
+        [UnitTest]
+        public void OnlyDataTablesInAssemblyPopulatesCorrectly()
+        {
+            var assembly = new FakeAssembly
+            {
+                TypesToReturn = new[]
+                {
+                    typeof(DateTime),
+                    typeof(StubDataTableOne),
+                    typeof(StubDataTableTwo),
+                    typeof(StubDataTableThree),
+                    typeof(Exception),
+                }
+            };
+
+            var serializer = new FakeSerializer();
+
+            Assert.IsTrue(
+                TableDescriptorFactory.TryCreate(
+                    typeof(StubDataTableOne),
+                    serializer,
+                    out TableDescriptor expectedDescriptor1));
+            Assert.IsTrue(
+                TableDescriptorFactory.TryCreate(
+                    typeof(StubDataTableTwo),
+                    serializer,
+                    out TableDescriptor expectedDescriptor2));
+            Assert.IsTrue(
+                TableDescriptorFactory.TryCreate(
+                    typeof(StubDataTableThree),
+                    serializer,
+                    out TableDescriptor expectedDescriptor3));
+
+            var sut = TableDiscovery.CreateForAssembly(assembly);
+            var tables = sut.Discover(new FakeSerializer());
+
+            Assert.AreEqual(3, tables.Count);
+            Assert.IsNotNull(tables.Single(x => x.Descriptor.Equals(expectedDescriptor1)));
+            Assert.IsNotNull(tables.Single(x => x.Descriptor.Equals(expectedDescriptor2)));
+            Assert.IsNotNull(tables.Single(x => x.Descriptor.Equals(expectedDescriptor3)));
+        }
+
+        [TestMethod]
+        [UnitTest]
+        public void OnlyMetadataTablesInAssemblyPopulatesCorrectly()
+        {
+            var assembly = new FakeAssembly
+            {
+                TypesToReturn = new[]
+                {
+                    typeof(DateTime),
+                    typeof(StubMetadataTableOne),
+                    typeof(StubMetadataTableTwo),
+                    typeof(Exception),
+                }
+            };
+
+            var serializer = new FakeSerializer();
+
+            Assert.IsTrue(
+                TableDescriptorFactory.TryCreate(
+                    typeof(StubMetadataTableOne),
+                    serializer,
+                    out TableDescriptor expectedDescriptor1));
+            Assert.IsTrue(
+                TableDescriptorFactory.TryCreate(
+                    typeof(StubMetadataTableTwo),
+                    serializer,
+                    out TableDescriptor expectedDescriptor2));
+
+            var sut = TableDiscovery.CreateForAssembly(assembly);
+            var tables = sut.Discover(new FakeSerializer());
+
+            Assert.AreEqual(2, tables.Count);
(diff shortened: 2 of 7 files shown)


documentation/Migration/Update-1.0.0-rc.md
```diff
@@ -9,6 +9,23 @@ into two sections: [Breaking Changes](#breaking-changes) and
...
+
+These parameters are being removed and replaced by the `ITableProvider` interface.
+If you have custom logic for determining the tables exposed by a `ProcessingSource`,
+you must implement the new interface.
+
+The default behavior of using all tables found in the assembly has been preserved.
+This change only effects those `ProcessingSource`s that use custom table providers.
+
+Two helper methods have been added for use:
+`TableDiscovery.CreateForAssembly` and `TableDiscovery.CreateForNamespace.` User's
```
mslukebo, line 26: typo: "User's" -> "Users"

zachnew-msft: Thanks!

src/Microsoft.Performance.SDK.Tests/TableDiscoveryTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            };
+
+            var serializer = new FakeSerializer();
+
+            Assert.IsTrue(
+                TableDescriptorFactory.TryCreate(
+                    typeof(StubMetadataTableOne),
+                    serializer,
+                    out TableDescriptor expectedDescriptor1));
+            Assert.IsTrue(
```
mslukebo, line 98: notice this is done a lot. Maybe we need a test util file that takes in a list of types and has out param for a list of descriptors?

src/Microsoft.Performance.SDK/Processing/ITableDiscoverer.cs
```diff
@@ -0,0 +1,309 @@
...
+namespace Microsoft.Performance.SDK.Processing
+{
+    /// <summary>
+    ///     Provides an interface for providing tables that are exposed
+    ///     by <see cref="ProcessingSource"/>s.
+    /// </summary>
+    public interface ITableProvider
+    {
+        /// <summary>
+        ///     Returns a collection of tables.
```
mslukebo, line 20: nit: this is a vague description. Maybe a "Returns the collection of tables that should be associated with a ProcessingSource" or something? Any thoughts?

src/Microsoft.Performance.SDK/Processing/ITableDiscoverer.cs
```diff
@@ -0,0 +1,309 @@
...
+        /// <returns>
+        ///     A collection of tables.
+        /// </returns>
+        ISet<DiscoveredTable> Discover(ISerializer tableConfigSerializer);
+    }
+
+    /// <summary>
+    ///     Provides common <see cref="ITableProvider"/> implementations.
+    /// </summary>
+    public static class TableDiscovery
```
mslukebo, line 34: can/should we put this in its own file?

src/Microsoft.Performance.SDK/Processing/ITableDiscoverer.cs
```diff
@@ -0,0 +1,309 @@
...
+        ///     <see cref="TableAttribute"/> attribute.
+        /// </summary>
+        /// <param name="assembly">
+        ///     The <see cref="Assembly"/> to search for tables.
+        /// </param>
+        /// <returns>
+        ///     A new <see cref="ITableProvider"/> that discovers tables in the
+        ///     specified <paramref name="assembly"/>.
+        /// </returns>
+        public static ITableProvider CreateForAssembly(Assembly assembly)
```
mslukebo, line 49: would it make sense to also have methods `CreateForAssembly(IProcessingSource)` and `CreateForNamespace(IProcessingSource)` that use the processing source's assembly/namespace? I could imagine that being a common scenario

src/Microsoft.Performance.SDK/Processing/ITableDiscoverer.cs
```diff
@@ -0,0 +1,309 @@
...
+                if (TableDescriptorFactory.TryCreate(t, tableConfigSerializer, out var isInternal, out var td, out var buildTable))
+                {
+                    s.Add(new DiscoveredTable(td, buildTable, isInternal));
+                }
+            }
+
+            return s;
+        }
+
+        private sealed class AssemblyTableDiscoverer
```
mslukebo, line 153: why are these classes private? What's the benefit of having the above static methods instead of these classes just having public constructors?

src/Microsoft.Performance.SDK/Processing/ITableDiscoverer.cs
```diff
@@ -0,0 +1,309 @@
...
+                    tableConfigSerializer);
+            }
+        }
+    }
+
+    /// <summary>
+    ///     Represents a table discovered by an <see cref="ITableProvider"/>
+    ///     instance.
+    /// </summary>
+    public sealed class DiscoveredTable
```
mslukebo, line 202: new file?

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
```diff
@@ -46,8 +45,30 @@ protected ProcessingSource()
             // so unfortunately we have to use an older idiom of every constructor calling an
             // init method so that we can pass the concrete type's assembly.
             //
-            this.tableAssemblyProvider = () => this.GetType().Assembly;
-            this.additionalTablesProvider = () => new Dictionary<TableDescriptor, Action<ITableBuilder, IDataExtensionRetrieval>>();
+
+            this.tableDiscoverer = TableDiscovery.CreateForAssembly(this.GetType().Assembly);
```
mslukebo, line 49: this ctor can just call `this(TableDiscovery.CreateForAssembly(this.GetType().Assembly)`, no?

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
```diff
@@ -95,6 +118,7 @@ protected ProcessingSource()
         ///     - or -
         ///     <paramref name="tableAssemblyProvider"/> is <c>null</c>.
         /// </exception>
+        [Obsolete("This constructor will be removed in 1.0.0-rc. Please use the constructor that takes an ITableProvider.")]
```
mslukebo, line 121: same here

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
zachnew-msft: you can't call this in this() or base(). I documented this in the other constructor, I can copy that documentation here as well.

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
```diff
@@ -62,6 +83,7 @@ protected ProcessingSource()
         /// <exception cref="System.ArgumentNullException">
         ///     <paramref name="additionalTablesProvider"/> is <c>null</c>.
         /// </exception>
+        [Obsolete("This constructor will be removed in 1.0.0-rc. Please use the constructor that takes an ITableDiscoverer.")]
```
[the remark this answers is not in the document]
zachnew-msft: Yes it can, I will delete it.

src/Microsoft.Performance.SDK.Tests/TableDiscoveryTests.cs
zachnew-msft: That makes sense, thanks.

src/Microsoft.Performance.SDK/Processing/ITableDiscoverer.cs
zachnew-msft: That sounds better, I will update.

src/Microsoft.Performance.SDK/Processing/ITableDiscoverer.cs
zachnew-msft: Yes, the class is big enough that I should do that.

src/Microsoft.Performance.SDK/Processing/ITableDiscoverer.cs
zachnew-msft: Yes, that is a good idea. I will add it.

src/Microsoft.Performance.SDK/Processing/ITableDiscoverer.cs
zachnew-msft: This allows the end user to depend only on the interfaces. I wanted to avoid having concrete types for this in case we needed to refactor implementations, or add additional functionality.

src/Microsoft.Performance.SDK/Processing/ITableDiscoverer.cs
```diff
@@ -0,0 +1,309 @@
...
+    {
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="DiscoveredTable"/>
+        ///     class, indicating that the table with the given <see cref="TableDescriptor"/>
+        ///     exists, and uses the default build action.
```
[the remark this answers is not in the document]
zachnew-msft: I will document this better. It looks for BuildTable.

src/Microsoft.Performance.SDK/Processing/ITableDiscoverer.cs
zachnew-msft: Yes, thanks

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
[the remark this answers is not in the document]
jmaxson-ms: This is a little confusing...

So this whole constructor is going away then? Do we need DeprecatedConstructorSupportProvider at all?

Or, is the constructor staying and we're removing the obsolete attribute? But that doesn't make sense to me...

src/Microsoft.Performance.SDK/Processing/DiscoveredTable.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            var other = obj as DiscoveredTable;
+            return other != null &&
+                   this.Descriptor.Equals(other.Descriptor);
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            return this.Descriptor.GetHashCode();
```
mslukebo, line 120: this hash code does not take into account `IsInternal` or `BuildTableAction`. Is that intentional?

src/Microsoft.Performance.SDK/Processing/DiscoveredTable.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(this, obj))
+            {
+                return true;
+            }
+
+            var other = obj as DiscoveredTable;
+            return other != null &&
```
mslukebo, line 113: same as hash code

zachnew-msft: See comment on hashcode.

Issue #53, opened 2021-03-08
Update creating a simple plugin to show how to include author information


mslukebo:
We do not currently document how to include "about info." The SQL plugin also should be updated have author info.

Issue #87, opened 2021-07-07
Update samples to target latest SDK


mslukebo:
1. We should have our samples use ProjectReferences instead of PackageReferences so it's always targeting whatever branch they're part of
2. Our samples need to be migrated to 0.109.xx
3. Our CI build pipeline for PRs should make sure these samples compile probably 😄

PR #137, opened 2021-10-04
User/zachnew msft/author

zachnew-msft:
Implements #53

### DIFF
--- a/samples/SimpleDataSource/SimpleCustomDataSource.cs
+++ b/samples/SimpleDataSource/SimpleCustomDataSource.cs
@@ -41,6 +41,65 @@ namespace SampleCustomDataSource
     {
         private IApplicationEnvironment applicationEnvironment;
 
+        //
+        // Provides information about this Custom Data Source, such as the author,
+        // a project link, licensing, etc. This information can be used by tools
+        // to display "About" information for your Custom Data Source. For example,
+        // Windows Performance Analyzer (WPA) uses this information for Help->About.
+        //
+
+        public override CustomDataSourceInfo GetAboutInfo()
+        {
+            return new CustomDataSourceInfo
+            {
+                //
+                // The copyright notice for this Custom Data Source.
+                //
+                CopyrightNotice = "Copyright 2021 Microsoft Corporation. All Rights Reserved.",
+
+                //
+                // The license under which this Custom Data Source may be used.
+                //
+                LicenseInfo = new LicenseInfo
+                {
+                    Name = "MIT",
+                    Text = "Please see the link for the full license text.",
+                    Uri = "https://github.com/microsoft/microsoft-performance-toolkit-sdk/blob/main/LICENSE.txt",
+                },
+
+                //
+                // A collection of the people or entities that own this Custom Data Source.
+                //
+                Owners = new[]
+                {
+                    new ContactInfo
+                    {
+                        Address = "1 Microsoft Way, Redmond, WA 98052",
+                        EmailAddresses = new[]
+                        {
+                            "noreply@microsoft.com",
+                        },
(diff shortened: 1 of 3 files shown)


samples/SimpleDataSource/SimpleCustomDataSource.cs
```diff
@@ -41,6 +41,65 @@ public class SimpleCustomDataSource
     {
         private IApplicationEnvironment applicationEnvironment;
 
+        //
+        // Provides information about this Custom Data Source, such as the author,
```
jmaxson-ms, line 45: >Custom Data Source [](http://example.com/codeflow?start=43&length=18)

Do we want to move away from Custom Data Source?

samples/SimpleDataSource/SimpleCustomDataSource.cs
```diff
@@ -41,6 +41,65 @@ public class SimpleCustomDataSource
...
+        // a project link, licensing, etc. This information can be used by tools
+        // to display "About" information for your Custom Data Source. For example,
+        // Windows Performance Analyzer (WPA) uses this information for Help->About.
+        //
+
+        public override CustomDataSourceInfo GetAboutInfo()
```
jmaxson-ms, line 51: >CustomDataSourceInfo [](http://example.com/codeflow?start=24&length=20)

I think this is being renamed to ProcessingSourceInfo. Is that class avalable yet?

samples/SimpleDataSource/SimpleCustomDataSource.cs
```diff
@@ -41,6 +41,65 @@ public class SimpleCustomDataSource
     {
         private IApplicationEnvironment applicationEnvironment;
 
+        //
+        // Provides information about this Custom Data Source, such as the author,
+        // a project link, licensing, etc. This information can be used by tools
+        // to display "About" information for your Custom Data Source. For example,
+        // Windows Performance Analyzer (WPA) uses this information for Help->About.
+        //
```
jmaxson-ms, line 49: Should we include this comment in the SqlCustomDataSource file as well?

samples/SimpleDataSource/SimpleCustomDataSource.cs
zachnew-msft: We are, this is consistent with what is currently there. I will update in a separate change.

samples/SimpleDataSource/SimpleCustomDataSource.cs
zachnew-msft: No, as the other sample don't have all of the other comments either, it would be out of place.

samples/SimpleDataSource/SimpleCustomDataSource.cs
zachnew-msft: https://github.com/microsoft/microsoft-performance-toolkit-sdk/issues/87

samples/SimpleDataSource/SimpleCustomDataSource.cs
zachnew-msft: https://github.com/microsoft/microsoft-performance-toolkit-sdk/issues/87

jmaxson-ms (approved):

samples/SimpleDataSource/SimpleCustomDataSource.cs
mslukebo, line 51: This should be ProcessingSourceInfo

mslukebo: Looks like this was addressed. Nevermind!

mslukebo (approved):

PR #158, opened 2021-11-03
Update ColumnRoles to use strings instead of enum

trgibeau:


src/Microsoft.Performance.SDK.Runtime/DTO/PreV1/TableConfiguration.cs
```diff
@@ -110,64 +110,65 @@ internal class TableConfiguration
...
                 var role = columnRolePair.Key;
                 var entry = columnRolePair.Value;
 
-                var newEntry = DTO.Enums.ColumnRole.Invalid;
+                string newEntry = null;
 
                 switch(role)
                 {
                     case ColumnRole.Invalid:
+                    case ColumnRole.CountColumnMetadata:
```
mslukebo, line 125: why is this special?

src/Microsoft.Performance.SDK.Runtime/DTO/PreV1/TableConfiguration.cs
```diff
@@ -110,64 +110,65 @@ internal class TableConfiguration
...
+                        break;
+
+
                     case ColumnRole.StartThreadId:
                     case ColumnRole.EndThreadId:
                     case ColumnRole.HierarchicalTimeTree:
                     case ColumnRole.WaitDuration:
                     case ColumnRole.WaitEndTime:
-                        newEntry = Enums.ColumnRole.Invalid;
+                        newEntry = role.ToString();
```
mslukebo, line 135: Can we add a comment here explaining why we're doing this and how we're no longer guaranteeing a column role is "known" by the driver?

src/Microsoft.Performance.SDK/Processing/ColumnRole.cs
```diff
@@ -4,55 +4,48 @@
 namespace Microsoft.Performance.SDK.Processing
 {
     /// <summary>
-    ///     Defines the role of a column in a graph.
+    ///     Defines known string values of column roles.
```
mslukebo, line 7: "Known" is a bit vague. Whomst is it known by? 😆 

What would you think of something like

Defines a common set of column roles that that are guaranteed to be successfully saved/restored in a serialized `<see cref="TableConfiguration" />`. Plugins may use any string value defined in this class to communicate column roles to any SDK driver. If a `<see cref="TableConfiguration" />` includes a column role value that is _not_ defined here, it is not guaranteed to 
1. Be de/serialized correctly
2. Be understood by an SDK driver

src/Microsoft.Performance.SDK/Processing/ColumnRole.cs
```diff
@@ -4,55 +4,48 @@
...
         /// </summary>
-        RecTop,
+        public const string RecTop = "RecTop";
 
         /// <summary>
-        ///     Indicates the Height of the Rec.
+        ///     Indicates the Height of the Rectangle.
         /// </summary>
-        RecHeight,
+        public const string RecHeight = "RecHeight";
```
mslukebo, line 44: do we still want to keep the rec stuff around?

src/Microsoft.Performance.SDK/Processing/TableConfiguration.cs
```diff
@@ -197,16 +197,10 @@ public IEnumerable<ColumnConfiguration> Columns
...
-        ///     - or -
-        ///     <paramref name="columnRole"/> is equal to <see cref="ColumnRole.CountColumnMetadata"/>.
-        /// </exception>
         /// <exception cref="System.InvalidOperationException">
         ///     <paramref name="column"/> is a metadata column (see <see cref="AllMetadataColumns"/>.
         /// </exception>
-        public void AddColumnRole(ColumnRole columnRole, ColumnConfiguration column)
+        public void AddColumnRole(string columnRole, ColumnConfiguration column)
         {
             Guard.NotNull(column, nameof(column));
```
mslukebo, line 205: should we add a Guard.NotNullOrWhitespace to columnRole? If we do we need to update the docstring too

src/Microsoft.Performance.SDK/Processing/TableConfiguration.cs
```diff
@@ -262,7 +237,7 @@ public void AddColumnRole(ColumnRole columnRole, Guid columnGuid)
         /// <param name="columnRole">
         ///     The role to be removed.
         /// </param>
-        public void RemoveColumnRole(ColumnRole columnRole)
+        public void RemoveColumnRole(string columnRole)
```
mslukebo, line 240: since we're here should we update this to return a `bool` and just return the .Remove below?

mslukebo:
@trgibeau please update the migration docs as well 😄

src/Microsoft.Performance.SDK/Processing/TableConfiguration.cs
```diff
@@ -78,9 +78,9 @@ public TableConfiguration(string name)
             this.Name = name;
 
             this.columnsRO = new ReadOnlyCollection<ColumnConfiguration>(EmptyArray<ColumnConfiguration>.Instance);
-            this.columnRoles = new Dictionary<ColumnRole, Guid>();
+            this.columnRoles = new Dictionary<string, Guid>();
```
jmaxson-ms, line 81: >Dictionary<string, Guid>(); [](http://example.com/codeflow?start=35&length=27)

I believe these are ordinal, case-sensitive by default. Is that what we want here?

[https://docs.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings](https://docs.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings)

src/Microsoft.Performance.SDK.Runtime/DTO/PreV1/TableConfiguration.cs
trgibeau: Yeah, its possible that they could be serialized, better to be safe then sorry

src/Microsoft.Performance.SDK.Runtime/DTO/PreV1/TableConfiguration.cs
```diff
@@ -110,64 +110,65 @@ internal class TableConfiguration
...
-                        break;
-
-                    case ColumnRole.CountColumnMetadata:
-                        newEntry = Enums.ColumnRole.CountColumnMetadata;
+                        newEntry = Processing.ColumnRole.RecWidth;
                         break;
                 }
 
-                if (newEntry != Enums.ColumnRole.Invalid)
+                if (!string.IsNullOrEmpty(newEntry))
```
jmaxson-ms, line 171: >IsNullOrEmpty [](http://example.com/codeflow?start=28&length=13)

will this ever be string.Empty? if so, would it ever be whitespace?

src/Microsoft.Performance.SDK/Processing/TableConfiguration.cs
trgibeau: Ah good catch! Will update!

trgibeau: Unless we want them to be case sensitive?

src/Microsoft.Performance.SDK/Processing/TableConfiguration.cs
trgibeau: We didn't do it before, not sure if it matters?

src/Microsoft.Performance.SDK/Processing/TableConfiguration.cs
jmaxson-ms: Yeah, I'm not sure. That's why I asked ;)

src/Microsoft.Performance.SDK.Runtime/DTO/PreV1/TableConfiguration.cs
trgibeau: It should never be empty, just null. 

I can update it to just check for null?

src/Microsoft.Performance.SDK/Processing/TableConfiguration.cs
mslukebo: just thought it'd be nice since we're currently just disregarding Remove's return value. Could be useful for error handling down the road.

src/Microsoft.Performance.SDK/Processing/TableConfiguration.cs
jmaxson-ms: I'm tempted to say that because we know these are being modified by hand, we should make it case insensitive. But I don't feel strongly.

mslukebo: Conceptually these are keys. Are keys typically case sensitive, or is there not really a "typical" case?

jmaxson-ms: I think it's scenario specific. Dictionary allows you to specify a comparer

src/Microsoft.Performance.SDK/Processing/TableConfiguration.cs
```diff
@@ -224,35 +223,21 @@ public void AddColumnRole(ColumnRole columnRole, ColumnConfiguration column)
         /// <param name="columnGuid">
         ///     The guid of column to place into the role.
         /// </param>
-        /// <exception cref="InvalidEnumArgumentException">
-        ///     <paramref name="columnRole"/> is not a valid member of the
-        ///     <see cref="ColumnRole"/> enumeration.
-        ///     - or -
-        ///     <paramref name="columnRole"/> is equal to <see cref="ColumnRole.CountColumnMetadata"/>.
+        /// <exception cref="ArgumentNullException">
+        ///     <paramref name="columnRole"/> is null.
```
jmaxson-ms, line 227: >null [](http://example.com/codeflow?start=49&length=4)

nit: <c>null</c>

same above.... if you're making other changes

jmaxson-ms (approved):
:shipit:

jmaxson-ms (approved):
:shipit:

mslukebo (approved):

PR #163, opened 2021-11-04
User/anramanath/aggregate progress

### LINKED ISSUE #154: Create a AggregateProgressTracker class
Abstract the callback in SourceProcessingSession.cs using IProgress<int>

anramanath:
closes #154


src/Microsoft.Performance.SDK.Runtime/ProgressReporter.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <summary>
+        ///		Creates IProgress object which will update a listener everytime it is updated.
+        ///		It is provided the number of components that must finish for total progress to be completed
+        /// </summary>
+        /// <param name="listener">Update this listener everytime this.currentProgress updates</param>
+        /// <param name="numberOfParts">Number of times that this.currentProgress will reach 100 due to subcomp</param>
+        public ProgressReporter(IProgress<int> listener, int numberOfParts = 1)
+        {
+            this.listener = listener;
+            this.numberOfParts = numberOfParts;
```
zachnew-msft, line 33: We should make sure this is not negative or zero.

PR #168, opened 2021-11-12
Fix incorrect parameter in 1.0.0-rc1 Engine Update Docs

thhous-msft:
EngineCreateInfo takes a ReadOnlyDataSourceSet, but the docs were passing in a DataSourceSet, and there is no implicit conversion


zachnew-msft (approved):

mslukebo (approved):
Thanks for fixing this!

PR #178, opened 2021-11-30
Allow loading assemblies without an assembly location

### LINKED ISSUE #177: Support Single File .NET Executables
We have a WPA plugin in MsQuic where we also have a tool that lets us process ETL files using the command line. I'm working on updating that plugin to the newest SDK, and also making the tool be a .NET 6 single file exe.  Creating the single file exe isn't working. After some debugging, I've realized this is because DataSourceSet only loads plugins in a directory, and doesn't have a way to load plugins from an Assembly type that is already loaded. Since the exe is a single file, there is no plugin to load from file, which make our processing sources not work. A way to explicitly pass in assemblies to be searched would solve this problem.

thhous-msft:
Closes #177 

There are cases where the assembly containing a plugin type might already be loaded. In this case, it would be nice to be able to add the type directly, rather than loaded from a folder. Additionally, there are cases where the plugin assembly might not even exist on disk (This is the case with single file executables). In this case, the type has to be loaded explicitly. There is a way to do this for a single plugin using an IAssemblyLoader, but that fails if the type doesn't have a location. This fixes, so the type can always be loaded.

### DIFF
--- a/src/Microsoft.Performance.SDK.Runtime/AssemblyTypeReference.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/AssemblyTypeReference.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Performance.SDK.Runtime
 
             this.type = type;
 
-            if (!this.Type.Assembly.IsDynamic && !string.IsNullOrWhiteSpace(this.Type.Assembly.Location))
+            if (!this.Type.Assembly.IsDynamic)
             {
                 this.assemblyPath = this.Type.Assembly.Location;
                 this.version = FileVersionInfo.GetVersionInfo(this.AssemblyPath).FileVersion;
--- a/src/Microsoft.Performance.SDK.Runtime/AssemblyTypeReference.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/AssemblyTypeReference.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Performance.SDK.Runtime
 
             this.type = type;
 
-            if (!this.Type.Assembly.IsDynamic)
+            if (!this.Type.Assembly.IsDynamic && !string.IsNullOrWhiteSpace(this.Type.Assembly.Location))
             {
                 this.assemblyPath = this.Type.Assembly.Location;
                 this.version = FileVersionInfo.GetVersionInfo(this.AssemblyPath).FileVersion;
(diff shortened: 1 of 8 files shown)


src/Microsoft.Performance.Toolkit.Engine/PluginSet.cs
```diff
@@ -258,6 +259,49 @@ public static PluginSet Load(string extensionDirectory)
...
+        /// <param name="assemblyLoader">
+        ///     The loader to use to load plugin assemblies. This parameter may be
+        ///     <c>null</c>. If this parameter is <c>null</c>, then the default
+        ///     loader will be used.
+        ///     <remarks>
+        ///         The default loader provides no isolation.
+        ///     </remarks>
+        /// </param>
+        /// <param name="extraAssemblies">
+        ///     A list of extra assemblies to attemt to directly load plugins from.
```
nibanks, line 282:
```suggestion
        ///     A list of extra assemblies to attempt to directly load plugins from.
```

mslukebo:
This change introduces a large shift in how we think about/interact with plugins, and it's not clear if we want to support loading plugins from executables. I think the current desire is to have plugins be a set of DLLs sitting in some directory.

nibanks:
Our (the MsQuic team) goal here is to have the ability to produce a single EXE binary that can be used to process ETL files, without any additional installation requirements. This change is required to achieve that goal as far as I can tell.

src/Microsoft.Performance.SDK.Runtime/AssemblyTypeReference.cs
```diff
@@ -63,7 +63,7 @@ protected internal AssemblyTypeReference(Type type)
 
         /// <summary>
         ///     Initializes a new instance of <see cref="AssemblyTypeReference{TDerived}"/> using
-        ///     <paramref name="other"/>'s <see cref="AssemblyTypeReference{TDerived}.Type"/>, 
+        ///     <paramref name="other"/>'s <see cref="AssemblyTypeReference{TDerived}.Type"/>,
```
nibanks, line 66: I assume your editor automatically did this, but probably best to revert this change.

nibanks: nvm, I was looking at the diff between commits. You're already touching the file.

thhous-msft: Changing editor settings sucks. Its not a big deal IMO.

nibanks (approved):

mslukebo (approved):

trgibeau (approved):

PR #209, opened 2022-09-08
ProcessingSource exposes its IProcessingSourceTableProvider to derived classes.

jmaxson-ms:
The abstract class, ProcessingSource, requires that, when provided, the IProcessSourceTableProvider be passed into the constructor. This does not provide the derived class an opportunity to initialize the IProcessSourceTableProvider object.

A new virtual method, CreateTableProvider, offers another approach. This method is called when the table provider is required, and a derived class may provide an implementation at this time by overriding this method. The default behavior remains the same.

This will require a minor version bump.

### DIFF
--- a/src/Microsoft.Performance.SDK.Tests/ProcessingSourceTests.cs
+++ b/src/Microsoft.Performance.SDK.Tests/ProcessingSourceTests.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Performance.SDK.Tests
         [TestCleanup]
         public void Cleanup()
         {
-            StubDataSource.Assembly = typeof(StubDataSource).Assembly;
+            AssemblyBasedStubDataSource.Assembly = typeof(AssemblyBasedStubDataSource).Assembly;
         }
 
         [TestMethod]
@@ -51,9 +51,9 @@ namespace Microsoft.Performance.SDK.Tests
                 typeof(StubDataTableOneNoBuildMethod),
                 typeof(StubDataTableTwoNoBuildMethod));
 
-            StubDataSource.Assembly = assembly;
+            AssemblyBasedStubDataSource.Assembly = assembly;
 
-            var sut = new StubDataSource();
+            var sut = new AssemblyBasedStubDataSource();
             sut.SetApplicationEnvironment(applicationEnvironment);
 
             Assert.AreEqual(expectedDescriptors.Count, sut.AllTablesExposed.Count);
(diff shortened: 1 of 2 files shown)


mslukebo:
Since the goal is to ensure a discoverer is initialized before Discovery happens, and that initialization needs information off of the processing source, I would prefer we just remove the parameter from the constructor and add a virtual `ProcessingSource.CreateTableDiscoverer()` call.

src/Microsoft.Performance.SDK.Tests/ProcessingSourceTests.cs
```diff
@@ -137,7 +137,9 @@ public void ManyDataSourcesWhenSubClassReturnsNullProcessorThenThrows()
 
         [TestMethod]
         [UnitTest]
-        public void WhenTableDiscoveryProvidedUsesDiscovery()
+        [DataRow(true)]  // todo: this datarow is temporary and should be removed once the deprecated
+        [DataRow(false)] // ProcessingSource.ProcessingSource(IProcessingSourceTableProvider) constructor is removed
```
mslukebo, line 141: Suggestion:

refactor this into 3 separate tests:

1. Assert that when using the ctor with no parameters, the expected default is returned from `CreateTableDiscoverer`
2. Assert that when the obsolete ctor is used, the same instance passed to the ctor is returned from `CreateTableDiscoverer`
3. Assert that the table returned from `CreateTableDiscover` is the one used to discover tables when tables are initialized

jmaxson-ms: gah, I was writing back to explain why this was all tested... when I realized that SubDataSource's default constructor doesn't test the base class' base constructor. thanks for pointing this out. I'll do a bigger update on this test as you suggest.

jmaxson-ms: ok, updated

src/Microsoft.Performance.SDK.Tests/ProcessingSourceTests.cs
```diff
@@ -170,6 +170,27 @@ public void WhenTableDiscoveryProvidedUsesDiscovery()
...
+        [UnitTest]
+        public void WhenTableDiscoveryNotProvided_DefaultUsedForDiscovery()
+        {
+            var sut = new DefaultTableProviderStubDataSource();
+            sut.SetApplicationEnvironment(applicationEnvironment);
+
+            // note: the table provider is assembly based so we can't easily check the contents in the unit test,
+            // but other tests check the contents, and the purpose of this test is just to check that a default
+            // provider was created.
+            Assert.IsNotNull(sut.TableProvider, "A default table provider should be used, but is null.");
```
mslukebo, line 191: We can

```cs
Assert.AreEqual(typeof(AssemblyTableDiscoverer), sut.TableProvider.GetType())
```
```

mslukebo (approved):

anramanath (approved):

PR #210, opened 2022-09-19
Add API for processing groups of plugin-defined DataSource groups

### LINKED ISSUE #213: Allow plugins to specify which data source combinations can be processed together
**Is your feature request an entirely new concept?**
Yes.

**Is your feature request related to an existing component?**
Yes: `ICustomDataProcessor`. Currently, processors are asked to process an `IEnumerable<IDataSource>`. This feature request builds on this API.

**Is your feature request related to a problem? Please describe.**
Yes. A processor is asked to process a collection of data sources, but it has no control over specifying if data sources _can_ be processed together. A plugin may not be able to process two data sources together, but instead only one at a time. If a user opens two data sources specific to this plugin, the plugin cannot advertise "do not give me both data sources in one call to ProcessAsync; instead call ProcessAsync once for each data source." The best the plugin can currently do is throw an exception or silently ignore data sources in these scenarios.

**Describe the solution you'd like**
A plugin can have its `IProcessingSource` implement a `IDataSourceGrouper` interface. For v1.1, this interface is optional; for v2, it's required.

The interface would allow a plugin to return `IDataSourceGroup` instances based on the data sources + processor options associated with the plugin. Each instance represents a valid grouping of data sources that the plugin can process together.

Specifically, it has the definition

```cs
public interface IDataSourceGrouper
{
    IReadOnlyCollection<IDataSourceGroup> Group(IEnumerable<IDataSource> dataSources, ProcessorOptions options);
}
```

where it is guaranteed that each data source is supported by the implementing `IProcessingSource`.

mslukebo:
Closes #213 

Adds the interface described in #214, but does not implement using it during Engine processing.


src/Microsoft.Performance.SDK.Runtime/ExecutionContext.cs
```diff
@@ -69,7 +70,7 @@ public sealed class ExecutionContext
         ///     Gets the <see cref="IDataSource"/>s that can be
```
mslukebo, line 70: update

src/Microsoft.Performance.SDK/Processing/DefaultProcessingMode.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Performance.SDK.Processing
+{
+    public class DefaultProcessingMode
+        : IProcessingMode
+    {
+        private static readonly Guid guid = new Guid("F97C3411-5DF5-4633-9EFD-470BEEBF40EF");
```
mslukebo, line 11: make public?

src/Microsoft.Performance.SDK/Processing/DefaultProcessingMode.cs
```diff
@@ -0,0 +1,25 @@
...
+        public Guid Guid => DefaultProcessingMode.guid;
+
+        /// <inheritdoc />
+        public string Name => "Default";
+
+        /// <inheritdoc />
+        public string Description => "A non-special method of processing data sources.";
+
+        /// <inheritdoc />
+        public bool SupportsGroupCombining => true;
```
mslukebo, line 23: remove

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
```diff
@@ -197,6 +198,46 @@ public virtual ProcessingSourceInfo GetAboutInfo()
...
+            IProcessorEnvironment processorEnvironment,
+            ProcessorOptions options)
+        {
+            if (this is IDataSourceGrouper)
+            {
+                throw new InvalidOperationException(
+                    $"Prior to V2, you must override this method if you implement {nameof(IDataSourceGrouper)}");
+            }
+            
+            this.Logger.Warn($"{this.GetType().Name} does not support processing user-specified processing modes - falling back to default processing.");
```
mslukebo, line 235: processing GROUPS

src/Microsoft.Performance.SDK/Processing/DefaultProcessingMode.cs
```diff
@@ -0,0 +1,28 @@
...
+using System;
+
+namespace Microsoft.Performance.SDK.Processing
+{
+    /// <summary>
+    ///     A default <see cref="IProcessingMode"/>.
+    /// </summary>
+    public class DefaultProcessingMode
+        : IProcessingMode
+    {
```
anramanath, line 13: Lets add a static getter here :

`public static readonly DefaultProcessingMode Default = new DefaultProcessingMode();`

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -994,11 +995,24 @@ private RuntimeExecutionResults ProcessCore()
                 {
                     try
                     {
+                        /*
+                         * TODO: for now, the engine ignores the groups returned by processing sources that implement
```
anramanath, line 999: Mind adding #214 in here to track?

src/Microsoft.Performance.SDK/Processing/DefaultProcessingMode.cs
mslukebo: I'd rather let plugins create their own `IProcessingMode` classes so that they can store extra information on the class if they need it. For instance, a plugin might want to store some metadata about _how_ a processing mode should be configured on the class.

With the suggest approach, that wouldn't be possible. Plugins would also have to check for the processing mode to use with Guid comparisons (i.e. `if (group.ProcessingMode.Guid == MyProcessingModeGuid)`) instead of a type check (i.e. `if (group.ProcessingMode is MyProcessingMode)`).

I am going to seal this class though.

mslukebo: I do not see the benefit of adding a `public static readonly DefaultProcessingMode Default` field. When/why would it be used?

src/Microsoft.Performance.SDK/Processing/DataSourceGrouping/IDataSourceGroup.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Performance.SDK.Processing.DataSourceGrouping
+{
+    /// <summary>
+    ///     A <see cref="IDataSource"/> collection that a <see cref="ICustomDataProcessor"/> can process together as
+    ///     a single, combined <see cref="IDataSource"/> in the specified <see cref="ProcessingMode"/>.
```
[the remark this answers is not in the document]
mslukebo: Essentially, 1 or more `IDataSource`s that can be processed together in a single call to `CustomDataProcessor.ProcessAsync`.

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
```diff
@@ -254,6 +279,43 @@ public virtual void DisposeProcessor(ICustomDataProcessor processor)
...
+        ///     to return <c>null</c> from this method.
+        /// </returns>
+        protected virtual ICustomDataProcessor CreateProcessorCore(
+            IDataSourceGroup dataSourceGroup,
+            IProcessorEnvironment processorEnvironment,
+            ProcessorOptions options)
+        {
+            if (this is IDataSourceGrouper)
+            {
+                throw new InvalidOperationException(
```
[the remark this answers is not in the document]
mslukebo: Updated the exception message

anramanath (approved):

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
```diff
@@ -254,6 +279,43 @@ public virtual void DisposeProcessor(ICustomDataProcessor processor)
...
+                    $"Prior to V2, you must override the {nameof(CreateProcessorCore)} which accepts a {nameof(IDataSourceGroup)} if you implement {nameof(IDataSourceGrouper)}");
```
jmaxson-ms, line 311: >if you implement [](http://example.com/codeflow?start=130&length=16)

nittiest nit of all nits (feel free to ignore): "if you implement" -> "when implementing"

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
```diff
@@ -254,6 +279,43 @@ public virtual void DisposeProcessor(ICustomDataProcessor processor)
...
+            }
+            
+            this.Logger.Warn($"{this.GetType().Name} does not support processing user-specified processing groups - falling back to default processing.");
+	
+            // Call v1 methods for now
+            return this.CreateProcessor(dataSourceGroup.DataSources, processorEnvironment, options);
```
jmaxson-ms, line 317: Should we throw if dataSouceGroup.ProcessingMode is not DefaultProcessingMode?

mslukebo: I'd prefer not to, since that would break backwards compatibility with plugins that don't implement this method if somehow that's the case. I think the only way for it to not be DefaultProcessingMode is if a driver/engine is doing something wrong. I'd rather the plugin silently use the V1 path in the case.

Perhaps a logger.Warn?

src/Microsoft.Performance.SDK/Processing/DataSourceGrouping/DataSourceGroup.cs
```diff
@@ -0,0 +1,38 @@
...
+        ///     The <see cref="IDataSource"/>s in this group. This parameter may not be <c>null</c>.
+        /// </param>
+        /// <param name="processingMode">
+        ///     The <see cref="IProcessingMode"/> that this group should be processed in. This parameter may not
+        ///     be <c>null</c>.
+        /// </param>
+        public DataSourceGroup(IReadOnlyCollection<IDataSource> dataSources, IProcessingMode processingMode)
+        {
+            Guard.NotNull(dataSources, nameof(dataSources));
+            Guard.NotNull(processingMode, nameof(processingMode));
```
jmaxson-ms, line 26: We haven't used the <exception></exception> doc comments as consistently as we should, but we should try. Since this Guard.NotNull throws, we should add them here. Can copy/paste other examples in code, such as Engine.Create.

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
```diff
@@ -254,6 +279,43 @@ public virtual void DisposeProcessor(ICustomDataProcessor processor)
...
+        ///     The environment for this specific processor instance.
+        /// </param>
+        /// <param name="options">
+        ///     The options to pass to the processor.
+        /// </param>
+        /// <returns>
+        ///     A new <see cref="ICustomDataProcessor"/>. It is an error
+        ///     to return <c>null</c> from this method.
+        /// </returns>
+        protected virtual ICustomDataProcessor CreateProcessorCore(
```
jmaxson-ms, line 303: >CreateProcessorCore [](http://example.com/codeflow?start=47&length=19)

We haven't used the <exception></exception> doc comments as consistently as we should, but we should try.

        /// <exception cref="InvalidOperationException">
        ///
        /// </exception>

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
```diff
@@ -197,6 +198,30 @@ public virtual ProcessingSourceInfo GetAboutInfo()
             return processor;
         }
 
+        /// <inheritdoc />
+        public ICustomDataProcessor CreateProcessor(
```
jmaxson-ms, line 202: >CreateProcessor [](http://example.com/codeflow?start=36&length=15)

We haven't used the <exception></exception> doc comments as consistently as we should, but we should try.

Looks like several possible exceptions here.

jmaxson-ms: FYI, we've typically put expected exception comments on interfaces where appropriate. I have mixed feelings about this because an implementation might throw those listed exceptions, or possibly others. But interfaces on learn.microsoft.com do list possible exceptions, so I'm guessing that's why we've done it.

anramanath (approved):

src/Microsoft.Performance.SDK.Runtime/ProcessingSourceReference.cs
```diff
@@ -319,8 +320,8 @@ public IEnumerable<ICustomDataProcessor> TrackedProcessors
         /// <summary>
         ///     Creates a new processor for processing the specified data sources.
```
jmaxson-ms, line 321: >data sources [](http://example.com/codeflow?start=69&length=12)

nit: data source group

mslukebo: thanks!

jmaxson-ms:
>     ///     <see cref="IDataSource"/>s that can be processed by the

nit:IDataSourceGroup

---
Refers to: src/Microsoft.Performance.SDK.Runtime/DataSessionSource.cs:15 in e85f28a. [](commit_id = e85f28a5c5d7baa2c14432f08b2c00edad7a51fc, deletion_comment = False)

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -954,7 +955,7 @@ private RuntimeExecutionResults ProcessCore()
                         new ProcessingError(
```
jmaxson-ms, line 955: >ProcessingError [](http://example.com/codeflow?start=28&length=15)

should ProcessingError contain the DataSourceGroup now? the processing mode might be relevant to the error?

mslukebo: I thought about that, but wanted to minimize the changes since this PR is already bigger than expected. Happy to open a new work item though 😄

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -994,11 +995,24 @@ private RuntimeExecutionResults ProcessCore()
...
+                         * TODO (#214): for now, the engine ignores the groups returned by processing sources that implement
+                         * IDataSourceGrouper. In the future, this needs to
+                         * 1. Check if the processing source implements IDataSourceGrouper (which it must after V2)
+                         * 2. If not, do what we're doing now and just create a new DataSourceGroup with the default
+                         *     processing mode
+                         * 3. If yes, use an IDataSourceGroupResolver to pick out which of the IDataSourceGroup(s)
+                         *    returned by the processing source should be used for processing.
+                         *      a. If the engine user does not supply a resolver, a sensible one should be used. E.g.
+                         *         one that picks groups that maximize the number of data sources processed without a
+                         *         data source being processed in more than 1 group
```
jmaxson-ms, line 1008: i'm not sure 3.a is true... may be something we have to discuss, but this behavior might lead to misleading data for the user.

it might be better to simply throw an exception and force the host process to provide a grouper in this case.

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
jmaxson-ms: whichever you prefer

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
mslukebo: Yeah, that was just an example. We will need to decide this before implementing #214

src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs
mslukebo: Not in this case, since the check of `is IDataSourceGrouper` failed above. If we're hitting this line, it's an `IProcessingSource` that doesn't specify any processing modes.

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
mslukebo: Made #216

jmaxson-ms (approved):

anramanath (approved):

PR #212, opened 2022-10-03
Add Processor Options

### LINKED ISSUE #176: Enable custom ProcessorOptions through the Engine.
The Engine always passes ProcessorOptions.Default when creating processors today. See Engine.CreateExecutors.  We need a way to pass in custom options.

anramanath:
Enables Passing of Processor Options per DataSourceGroup to ProcessingSources. Closes #176


src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -38,6 +44,50 @@ public EngineCreateInfo(ReadOnlyDataSourceSet dataSources)
...
+        }
+
+        /// <summary>
+        ///     Set <see cref="ProcessorOptions"/> per <see cref="IProcessingSource"/>.
+        /// </summary>
+        /// <param name="processingSourceOptionsMap">A map of <see cref="IProcessingSource"/>s with <see cref="ProcessorOptions"/> to pass it.</param>
+        /// <exception cref="ArgumentNullException">
+        ///     <paramref name="processingSourceOptionsMap"/> is <c>null</c>.
+        /// </exception>
+        public void WithProcessorOptions(IDictionary<IProcessingSource, ProcessorOptions> processingSourceOptionsMap)
```
mslukebo, line 71: How does someone calling this method get the `IProcessingSource` instance(s) that would be in this dictionary?

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -67,6 +117,11 @@ public EngineCreateInfo(ReadOnlyDataSourceSet dataSources)
         /// </summary>
         public ReadOnlyDataSourceSet DataSources { get; }
 
+        /// <summary>
+        ///
```
mslukebo, line 121: Missing content

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -1008,14 +1024,22 @@ private RuntimeExecutionResults ProcessCore()
                          *         data source being processed in more than 1 group
                          * 4. Add a new executor for each IDataSourceGroups picked out in the above step
                          */
+
+                        var dsg = new DataSourceGroup(dataSources, new DefaultProcessingMode());
+                        var processorOptions = optionsResolver.GetProcessorOptions(dsg, processingSource.Instance);
+                        if (!processorOptions.Options.All(o => processingSource.CommandLineOptions.Any(clo => clo.Id.Equals((o.Id as Option)?.Id))))
+                        {
+                            throw new NotSupportedException($"All ProcessorOptions are not supported.");
```
mslukebo, line 1032: Couple of comments on this block:
1. `NotSupportedException` is probably not the most appropriate exception to throw here. `InvalidOperationException` would probably be closer. But...
2. Should we make a custom exception here?
3. "All ProcessorOptions are not supported" is a little vague. It would be more helpful to know exactly what the invalid option is, as well as what processing source is being used

anramanath: Hmmm, I agree with 2, for making a custom exception and I'll make it so the invalid options can be returned.

src/Microsoft.Performance.Toolkit.Engine.Tests/ToolkitEngineCreateInfoTests.cs
```diff
@@ -17,6 +29,8 @@ public void DefaultRuntime_Set()
 
             string expectedName = typeof(EngineCreateInfo).Assembly.GetName().Name;
             Assert.AreEqual(expectedName, sut.RuntimeName);
+            Assert.IsNotNull(sut.OptionsResolver, "Options Resolver is null when a default is expected");
```
mslukebo, line 32: Is there a reason to put this here instead of in its own unit test? The test this was added to is named "DefaultRuntime_Set," so I would expect it to _only_ fail if the default runtime is not set.

src/Microsoft.Performance.Toolkit.Engine.Tests/ToolkitEngineCreateInfoTests.cs
```diff
@@ -1,8 +1,20 @@
...
 // Licensed under the MIT License.
 
+using Microsoft.Performance.SDK.Processing;
+using Microsoft.Performance.SDK.Processing.DataSourceGrouping;
+using Microsoft.Performance.SDK.Runtime;
 using Microsoft.Performance.Testing;
+using Microsoft.Performance.Testing.SDK;
+using Microsoft.Performance.Toolkit.Engine.Tests.TestCookers.Source123;
+using Microsoft.Performance.Toolkit.Engine.Tests.TestCookers.Source4;
+using Microsoft.Performance.Toolkit.Engine.Tests.TestCookers.Source5;
```
mslukebo, line 11: Are these imports still needed?

src/Microsoft.Performance.Toolkit.Engine.Tests/ToolkitEngineCreateInfoTests.cs
anramanath: Haha, I didn't understand why the test was named that way. Going to make my own unit tests in EngineCreateInfo and Engine now.

src/Microsoft.Performance.SDK/Processing/ProcessorOptions.cs
```diff
@@ -64,5 +66,27 @@ public class ProcessorOptions
         ///     passed to the processor.
         /// </summary>
         public IEnumerable<string> Arguments { get; }
+
+        public override bool Equals(object obj)
```
mslukebo, line 70: Why do we need to override this and HashCode?

src/Microsoft.Performance.SDK/Processing/Options/GlobalProcessingOptionsResolver.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Performance.SDK.Processing.DataSourceGrouping;
+using System;
+using System.Linq;
+
+namespace Microsoft.Performance.SDK.Processing.Options
+{
+    public sealed class GlobalProcessingOptionsResolver : IProcessingOptionsResolver
```
mslukebo, line 10: This should probably be internal

src/Microsoft.Performance.SDK/Processing/Options/IProcessingOptionsResolver.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Performance.SDK.Processing.DataSourceGrouping;
+using System.Collections.Generic;
+
+namespace Microsoft.Performance.SDK.Processing.Options
+{
+    public interface IProcessingOptionsResolver
```
mslukebo, line 9: missing docstring

src/Microsoft.Performance.SDK/Processing/Options/IProcessingOptionsResolver.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        /// <summary>
+        ///     Get the processor options for a group of data sources processed by a given processor.
+        /// </summary>
+        /// <param name="dataSourceGroup"> A group of data sources </param>
+        /// <param name="processingSource"> a processing source to process the data sources. </param>
+        /// <returns>Options <see cref="ProcessorOptions"/> to pass to the <see cref="IProcessingSource"/>.</returns>
+        ProcessorOptions GetProcessorOptions(IDataSourceGroup dataSourceGroup, IProcessingSource processingSource);
```
mslukebo, line 17: I'm not sure passing `IProcessingSource` is going to work here, especially since `IProcessingSource` does not expose a `Guid`. We also don't want the implementors to mess around with the `IProcessingSource` instances that are being used by the engine to process things. This is both exposing too much and not enough at the same time 😆 

The [IDataSourceGroupResolver](https://github.com/microsoft/microsoft-performance-toolkit-sdk/blob/65348731b57f10e0d8d60b44f236676f3ce30d4d/src/Microsoft.Performance.SDK/Processing/DataSourceGrouping/IDataSourceGroupResolver.cs) interface passes in the `Guid` of the processing source. For consistency, we can do that.

Also for consistency, we should keep the position of the parameters the same (i.e. both have the guid first or second). Do you mind updating either this or `IDataSourceGroupResolver` so that they match?

src/Microsoft.Performance.SDK/Processing/OptionInstance.cs
```diff
@@ -77,5 +77,25 @@ public OptionInstance(object optionId, IEnumerable<string> arguments)
         ///     Gets the arguments provided to this option.
         /// </summary>
         public IEnumerable<string> Arguments { get; }
+
+        public override bool Equals(object obj)
```
mslukebo, line 81: Do we need to override this and GetHashCode?

src/Microsoft.Performance.SDK/Processing/Option.cs
```diff
@@ -141,6 +141,38 @@ public class Option
         /// </summary>
         public int MaximumParameterCount { get; }
 
+        public override bool Equals(object obj)
```
mslukebo, line 144: Same here

src/Microsoft.Performance.SDK/Processing/Options/ProcessingSourceOptionsResolver.cs
```diff
@@ -0,0 +1,71 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.Performance.SDK.Processing.DataSourceGrouping;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Performance.SDK.Processing.Options
+{
+    public sealed class ProcessingSourceOptionsResolver : IProcessingOptionsResolver
```
mslukebo, line 12: This can be internal

src/Microsoft.Performance.SDK/Processing/Options/ProcessingSourceOptionsResolver.cs
```diff
@@ -0,0 +1,71 @@
...
+            Guard.NotNull(processingSource, nameof(processingSource));
+
+            var success = processingSourceToOptionsMap.TryGetValue(processingSource, out ProcessorOptions processorOptions);
+            if (!success || processorOptions == null)
+            {
+                processorOptions = ProcessorOptions.Default;
+            }
+
+            // Ensure that the ProcessingSource supports all provided options
+            if (!processorOptions.Options.All(o => processingSource.CommandLineOptions.Any(clo => clo.Id.Equals((o.Id as Option)?.Id))))
```
mslukebo, line 63: This can be removed

src/Microsoft.Performance.SDK.Tests/ProcessingOptionsResolverTests.cs
```diff
@@ -0,0 +1,303 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Performance.SDK.Tests
+{
+    [TestClass]
+    public class ProcessingOptionsResolverTests
+    {
+        [TestMethod]
+        [IntegrationTest]
```
mslukebo, line 23: I don't think this is an integration test since it appears to be testing the spec of `GlobalProcessingOptionsResolver` claiming "Each call to GetProcessorOptions returns the given value."

src/Microsoft.Performance.SDK.Tests/ProcessingOptionsResolverTests.cs
```diff
@@ -0,0 +1,303 @@
...
+        public void DefaultGlobalProcessingOptionsResolverTest()
+        {
+            var sut = GlobalProcessingOptionsResolver.Default;
```
mslukebo, line 26: This test can be dramatically simplified by removing the `GlobalProcessingOptionsResolver.Default` static field and just having the `EngineCreateInfo` ctor do `new GlobalProcessingOptionsResolver(ProcessorOptions.Default)`. Then, the only test you need for the entire class is "Create a GlobalProcessingOptionsResolver with a specified option. Assert that option is always returned when GetProcessorOptions is called".

src/Microsoft.Performance.SDK.Tests/ProcessingOptionsResolverTests.cs
```diff
@@ -0,0 +1,303 @@
...
+
+            var sut = new GlobalProcessingOptionsResolver(processorOptions);
+
+            Assert.IsNotNull(sut, "Options Resolver is null");
+
+            ProcessingSourceReference.TryCreateReference(typeof(Source123DataSource), out var psr123);
+            ProcessingSourceReference.TryCreateReference(typeof(Source4DataSource), out var psr4);
+
+            // success
+            Assert.AreEqual(processorOptions, sut.GetProcessorOptions(dsg4, psr4.Instance), "ProcessorOptions differ from expected");
```
mslukebo, line 129: these two cases should be entirely separate tests

src/Microsoft.Performance.SDK/Processing/IProcessingOptionsResolver.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.Performance.SDK.Processing
+{
+    public interface IProcessingOptionsResolver
+    {
+        /// <summary>
+        ///     Get the processor options for a group of data sources processed by a given processor.
+        /// </summary>
+        /// <param name="dataSourceGroup"> A group of data sources </param>
+        /// <param name="processingSource"> a processing source to process the data sources. </param>
```
jmaxson-ms, line 15: nit: add a period at the end.

Also, cap 'A' on next line.

src/Microsoft.Performance.SDK/Processing/ProcessorOptions.cs
```diff
@@ -1,15 +1,17 @@
...
+using System.Diagnostics;
 using System.Linq;
 
 namespace Microsoft.Performance.SDK.Processing
 {
     /// <summary>
     ///     Provides a means of passing options to a data processor.
     /// </summary>
-    public class ProcessorOptions
+    public class ProcessorOptions
```
jmaxson-ms, line 14: >  [](http://example.com/codeflow?start=33&length=1)

nit: cleanup extra space

src/Microsoft.Performance.Toolkit.Engine/DataSourceSet.cs
```diff
@@ -390,6 +396,7 @@ public bool TryAddDataSources(IEnumerable<IDataSource> dataSources, Type process
                 return false;
             }
 
+
             try
```
jmaxson-ms, line 400: nit: remove extra line

jmaxson-ms:
>                 freeDataSources,

this.

---
Refers to: src/Microsoft.Performance.Toolkit.Engine/DataSourceSet.cs:453 in f7e5056. [](commit_id = f7e5056009146dff872fafa7244ccb88b8b49648, deletion_comment = False)

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -287,6 +287,9 @@ public static Engine Create(IDataSource dataSource)
         /// <exception cref="ObjectDisposedException">
         ///     This instance is disposed.
         /// </exception>
+        /// <exception cref="NotSupportedException">
+        ///     The <c>createInfo.OptionsResolver.</c> are not supported by the provided Processing Source.
+        /// </exception>
         public static Engine Create(
```
jmaxson-ms, line 293: If this is generating a default EngineCreateInfo, this should never be thrown, should it?

Same for the next method.

src/Microsoft.Performance.Toolkit.Engine/ProcessingSourceOptionsResolver.cs
```diff
@@ -0,0 +1,72 @@
...
+using Microsoft.Performance.SDK;
+using Microsoft.Performance.SDK.Processing;
+using Microsoft.Performance.SDK.Processing.DataSourceGrouping;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Performance.Toolkit.Engine
+{
+    internal sealed class ProcessingSourceOptionsResolver : IProcessingOptionsResolver
```
jmaxson-ms, line 13: >: IProcessingOptionsResolver [](http://example.com/codeflow?start=58&length=28)

we place these on the next line

src/Microsoft.Performance.Toolkit.Engine/GlobalProcessingOptionsResolver.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Performance.SDK.Processing.DataSourceGrouping;
+using System;
+using System.Linq;
+
+namespace Microsoft.Performance.SDK.Processing.Options
+{
+    internal class GlobalProcessingOptionsResolver : IProcessingOptionsResolver
```
jmaxson-ms, line 10: >: IProcessingOptionsResolver [](http://example.com/codeflow?start=51&length=28)

we place these on the next line

src/Microsoft.Performance.SDK/Processing/ProcessorOptions.cs
anramanath: We need to be able to compare our ProcessorOptions in tests to the expected options.

src/Microsoft.Performance.SDK/Processing/OptionInstance.cs
anramanath: Yep! To be able to compare OptionInstances via ProcessorOptions for testing.

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
anramanath: Good catch :). Thanks.

src/Microsoft.Performance.SDK.Tests/ProcessingOptionsResolverTests.cs
anramanath: If it's explicitly using the new `GlobalProcessingOptionsResolver(ProcessorOptions.Default)` or not I think this redundant test is good to have since it'll be a common code path.

src/Microsoft.Performance.Toolkit.Engine.Tests/ProcessingOptionsResolverTests.cs
```diff
@@ -0,0 +1,254 @@
...
+
+            var dsg2 = new DataSourceGroup(
+                new[] { new FileDataSource("test2" + Source4DataSource.Extension) },
+                new DefaultProcessingMode());
+
+            ProcessingSourceReference.TryCreateReference(typeof(Source123DataSource), out var psr123);
+            ProcessingSourceReference.TryCreateReference(typeof(Source4DataSource), out var psr4);
+
+            // Ensure these are default
+            Assert.AreEqual(ProcessorOptions.Default, sut.GetProcessorOptions(psr123.Guid, dsg1), "ProcessorOptions differ from expected Default");
```
mslukebo, line 42: these assertions are duplicated many times in this file. I would make a helper method that takes in
1. The sut
2. The expected processor options
3. A list of PSR/DSG pairs

and the method can enumerate the list, asserting the expected options are returned from the call.

src/Microsoft.Performance.Toolkit.Engine.Tests/ProcessingOptionsResolverTests.cs
```diff
@@ -0,0 +1,254 @@
...
+    {
+        [TestMethod]
+        [UnitTest]
+        public void DefaultGlobalProcessingOptionsResolverTest()
+        {
+            var sut = new GlobalProcessingOptionsResolver(ProcessorOptions.Default);
+
+            Assert.IsNotNull(sut, "Options Resolver is null");
+
+            var dsg1 = new DataSourceGroup(
```
mslukebo: There is a method `Any.DataSourceGroup()` you can call to quickly create these instances.

src/Microsoft.Performance.Toolkit.Engine.Tests/ProcessingOptionsResolverTests.cs
```diff
@@ -0,0 +1,254 @@
...
+
+            Assert.AreEqual(processorOptions[0], sut.GetProcessorOptions(psr123.Guid, dsg123_1), "ProcessorOptions differ from expected");
+            Assert.AreEqual(processorOptions[0], sut.GetProcessorOptions(psr123.Guid, dsg123_2), "ProcessorOptions differ from expected");
+            Assert.AreEqual(processorOptions[1], sut.GetProcessorOptions(psr4.Guid, dsg4), "ProcessorOptions differ from expected");
+            Assert.AreEqual(processorOptions[0], sut.GetProcessorOptions(psr5.Guid, dsg5), "ProcessorOptions differ from expected");
+        }
+
+        [TestMethod]
+        [UnitTest]
+        public void CustomProcessingOptionsResolver()
```
mslukebo, line 166: this, and the mock resolver, can be removed. You're essentially testing your implementation of a mock, which does not add value to the suite of unit tests, since the mock isn't used by consumers of the SDK.

mslukebo (changes requested):

src/Microsoft.Performance.SDK/Processing/ProcessorOptions.cs
mslukebo: I don't think that's true; see my comments on the tests.

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -27,6 +29,9 @@ static EngineCreateInfo()
         /// <summary>
         ///     Initializes a new instance of the <see cref="EngineCreateInfo"/> class.
         /// </summary>
+        /// <remarks>
+        ///     Uses default ProcessorOptions. To modify this behavior <see cref="WithProcessorOptions(IProcessingOptionsResolver)"/>.
```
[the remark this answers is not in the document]
anramanath: Hmm, the goal here was to point out that we _can_ modify the behavior to pass in options and how to do so. Would you prefer to delete the remarks all together?

src/Microsoft.Performance.Toolkit.Engine.Tests/ProcessingOptionsResolverTests.cs
anramanath: If I were to use Any.DataSourceGroup it looks like it would always create the same dsg. Which is not great for all tests.

src/Microsoft.Performance.Toolkit.Engine.Tests/ProcessingOptionsResolverTests.cs
```diff
@@ -0,0 +1,254 @@
...
+namespace Microsoft.Performance.SDK.Tests
+{
+    [TestClass]
+    public class ProcessingOptionsResolverTests
+    {
+        [TestMethod]
+        [UnitTest]
+        public void DefaultGlobalProcessingOptionsResolverTest()
+        {
+            var sut = new GlobalProcessingOptionsResolver(ProcessorOptions.Default);
```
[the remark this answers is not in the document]
anramanath: This will _only_ work for ProcessorOptions.Default. Overriding ProcessorOptions.Equals() is fine.

src/Microsoft.Performance.Toolkit.Engine.Tests/ProcessingOptionsResolverTests.cs
```diff
@@ -0,0 +1,254 @@
...
+                new[]
+                {
+                    new FileDataSource("test5" + Source5DataSource.Extension),
+                }, new DefaultProcessingMode());
+
+            ProcessingSourceReference.TryCreateReference(typeof(Source123DataSource), out var psr123);
+            ProcessingSourceReference.TryCreateReference(typeof(Source4DataSource), out var psr4);
+            ProcessingSourceReference.TryCreateReference(typeof(Source5DataSource), out var psr5);
+
+            Assert.AreEqual(processorOptions, sut.GetProcessorOptions(psr123.Guid, dsg123), "ProcessorOptions differ from expected");
```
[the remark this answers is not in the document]
anramanath: This doesn't require it. Agreed. 

Test line 201 and 202 require it. I think it's fine to override the equals method. Why should we not override it?

src/Microsoft.Performance.Toolkit.Engine.Tests/ProcessingOptionsResolverTests.cs
anramanath: I think it _can_ be useful to have a mock options resolver which is already tested for future tests which might want to check specific behaviors for options resolvers, in case customers have issues with specific code paths.

Feel free to unresolve if you feel strongly about deleting this test case.

mslukebo: I'd prefer getting rid of it. Unless the mock is both very complicated and used actually test collateral, there's no reason to have unit tests for it.

src/Microsoft.Performance.Toolkit.Engine.Tests/ProcessingOptionsResolverTests.cs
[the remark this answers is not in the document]
mslukebo: It should not require it according to the spec of the systems under test. You should be able to modify your tests to compare the returned instances to the ones given to the ctors of the SUTs.

Overriding Equals and GetHashCode can lead to many subtle bugs. This PR does not introduce tests for Equals/GetHashCode for the 3 classes whose had those methods overridden. Unless the spec for a class requires overriding Equals, it's preferable to leave them alone.

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -983,59 +983,130 @@ private RuntimeExecutionResults ProcessCore()
...
+            ILogger logger)
+        {
+            bool areUnsupportedOptions = false;
+            foreach (var option in processorOptions.Options)
+            {
+                // Compare Option to every supported Option supported by the processingSourceReference
+                if (!processingSourceReference.CommandLineOptions.Any(clo => clo.Id.Equals(((Option)option.Id).Id)))
+                {
+                    areUnsupportedOptions = true;
+                    logger.Error($"{option} is unsupported");
```
mslukebo, line 1103: It might be helpful here to log information about the processing source, so the user can see what PSR doesn't support the given option

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -983,59 +983,130 @@ private RuntimeExecutionResults ProcessCore()
...
-                           "Error processing dataSources {0} on processing source {1}: {2}",
-                           dataSources.ToString(),
+                           "Unable to get processor options for processing source {0} and data sources {1}: {2}",
                            processingSource.Name,
+                           dsg.DataSources.ToString(),
                            e);
                     }
+
+                    // Validate processor options
+                    Debug.Assert(processorOptions != null, "How did we get here");
```
mslukebo, line 1021: We can easily get here if the user-implemented resolver returns null.  Perhaps we should just check for null in `AreUnsupportedOptions`?

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -38,6 +39,52 @@ public EngineCreateInfo(ReadOnlyDataSourceSet dataSources)
             Guard.NotNull(dataSources, nameof(dataSources));
 
             this.DataSources = dataSources;
+            this.OptionsResolver = new GlobalProcessingOptionsResolver(ProcessorOptions.Default);
+        }
+
+        /// <summary>
+        ///     Specifies an <see cref="IProcessingOptionsResolver"/> to use when processing data sources.
```
mslukebo, line 46: This method signature (and the one below) does not specify the resolver. The summary for each method and the docstring for the parameter should probably be almost identical. E.g. this should simply state "Specifies the ProcessorOptions to use..."

src/Microsoft.Performance.Toolkit.Engine/EngineException.cs
```diff
@@ -2,6 +2,8 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
```
mslukebo, line 6: These can be removed now

src/Microsoft.Performance.SDK/Utils.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Performance.SDK
+{
+    internal class Utils
+    {
+        internal static bool EnumerableComparer<T>(IEnumerable<T> list1, IEnumerable<T> list2)
```
mslukebo, line 10: Do we still need this method/file?

src/Microsoft.Performance.Toolkit.Engine.Tests/TestCookers/Source123/Source123DataSource.cs
```diff
@@ -44,6 +51,8 @@ protected override bool IsDataSourceSupportedCore(IDataSource dataSource)
                    Path.GetExtension(dataSource.Uri.LocalPath));
         }
 
+        public override IEnumerable<Option> CommandLineOptions => options;
```
mslukebo, line 54: Now that we don't directly use these PSRs in the unit tests, can all these be removed?

They could be used for integration testing the resolver if we want to do that, though.

mslukebo: Nevermind, didn't see the engine create info tests 😄

src/Microsoft.Performance.Toolkit.Engine.Tests/ToolkitEngineCreateInfoTests.cs
```diff
@@ -18,5 +24,75 @@ public void DefaultRuntime_Set()
...
+        {
+            var processorOptions = new ProcessorOptions(
+                new[]
+                {
+                    new OptionInstance(
+                        new Option('s', "test1"),
+                        "arg1"),
+                });
+
+            ProcessingSourceReference.TryCreateReference(typeof(Source123DataSource), out var psr);
```
mslukebo, line 73: I believe we can avoid creating the PSRs in this file just like the other test file, right?

src/Microsoft.Performance.Toolkit.Engine.Tests/ToolkitEngineCreateInfoTests.cs
```diff
@@ -18,5 +24,75 @@ public void DefaultRuntime_Set()
...
+            var sut = new EngineCreateInfo(DataSourceSet.Create().AsReadOnly());
+            sut.WithProcessorOptions(processorOptionsMap);
+
+            Assert.IsNotNull(sut.OptionsResolver, "Options Resolver is null when a default is expected");
+            Assert.AreEqual(typeof(ProcessingSourceOptionsResolver), sut.OptionsResolver.GetType(), "Options Resolver is not of type ProcessingSourceOptionsResolver");
+            Assert.AreEqual(processorOptions, sut.OptionsResolver.GetProcessorOptions(psr.Guid, dsg), "Expected processor options were not returned");
+        }
+
+        [TestMethod]
+        [IntegrationTest]
```
mslukebo, line 90: This is actually a unit test since it's checking the spec of a single method

src/Microsoft.Performance.Toolkit.Engine.Tests/ToolkitEngineCreateInfoTests.cs
```diff
@@ -18,5 +24,75 @@ public void DefaultRuntime_Set()
...
             Assert.AreEqual(expectedName, sut.RuntimeName);
         }
+
+        [TestMethod]
+        [IntegrationTest]
+        public void DefaultProcessorOptionsResolver_Set()
+        {
+            var sut = new EngineCreateInfo(DataSourceSet.Create().AsReadOnly());
+            Assert.IsNotNull(sut.OptionsResolver, "Options Resolver is null when a default is expected");
+            Assert.AreEqual(typeof(GlobalProcessingOptionsResolver), sut.OptionsResolver.GetType(), "Options Resolver is not of type GlobalProcessingOptionsResolver");
```
mslukebo, line 34: This is very glass-box test that's asserting implementation details instead of the spec.

Perhaps a more robust way to check this is to verify that `sut.OptionsResolver.GetProcessorOptions` returns `ProcessorOptions.Default`? Since that field is static, you can do reference equality on it against the returned options

src/Microsoft.Performance.Toolkit.Engine.Tests/ToolkitEngineCreateInfoTests.cs
```diff
@@ -18,5 +24,75 @@ public void DefaultRuntime_Set()
...
+                });
+
+            var dsg = new DataSourceGroup(new[] { new FileDataSource("sample") }, new DefaultProcessingMode());
+            ProcessingSourceReference.TryCreateReference(typeof(Source123DataSource), out var ps);
+
+            var sut = new EngineCreateInfo(DataSourceSet.Create().AsReadOnly());
+            sut.WithProcessorOptions(processorOptions);
+
+            Assert.IsNotNull(sut.OptionsResolver, "Options Resolver is null when a default is expected");
+            Assert.AreEqual(typeof(GlobalProcessingOptionsResolver), sut.OptionsResolver.GetType(), "Options Resolver is not of type GlobalProcessingOptionsResolver");
```
mslukebo, line 56: This is an implementation detail and shouldn't be asserted here

src/Microsoft.Performance.Toolkit.Engine.Tests/ToolkitEngineCreateInfoTests.cs
```diff
@@ -18,5 +24,75 @@ public void DefaultRuntime_Set()
...
+                { psr.Guid, processorOptions }
+            };
+
+            var dsg = new DataSourceGroup(new[] { new FileDataSource("sample") }, new DefaultProcessingMode());
+
+            var sut = new EngineCreateInfo(DataSourceSet.Create().AsReadOnly());
+            sut.WithProcessorOptions(processorOptionsMap);
+
+            Assert.IsNotNull(sut.OptionsResolver, "Options Resolver is null when a default is expected");
+            Assert.AreEqual(typeof(ProcessingSourceOptionsResolver), sut.OptionsResolver.GetType(), "Options Resolver is not of type ProcessingSourceOptionsResolver");
```
mslukebo, line 85: Same here

src/Microsoft.Performance.SDK/Processing/IProcessingOptionsResolver.cs
```diff
@@ -0,0 +1,24 @@
...
+    ///     Responsible for choosing the <see cref="ProcessorOptions"/> that will be used for processing.
+    /// </summary>
+    public interface IProcessingOptionsResolver
+    {
+        /// <summary>
+        ///     Get the processor options for a group of data sources processed by a given processor.
+        /// </summary>
+        /// <param name="processingSourceGuid"> Guid for processing source to process the data sources. </param>
+        /// <param name="dataSourceGroup"> A group of data sources. </param>
+        /// <returns> Options for processing.</returns>
```
mslukebo, line 20: This return value could be much more explicit. Something like
"The `<see cref="ProcessorOptions"/>` to associate with the `<see cref="IProcessingSource"/>` with the given `<paramref name="processingSourceGuid"/>`."

src/Microsoft.Performance.SDK/Processing/IProcessingOptionsResolver.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Microsoft.Performance.SDK.Processing
+{
+    /// <summary>
+    ///     Responsible for choosing the <see cref="ProcessorOptions"/> that will be used for processing.
+    /// </summary>
+    public interface IProcessingOptionsResolver
+    {
+        /// <summary>
+        ///     Get the processor options for a group of data sources processed by a given processor.
```
mslukebo, line 16: The summary and param tags in this docstring can use `see` tags in multiple places

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
anramanath: I added this assert to double check that options = null is what the user expected since it's a unique case that likely shouldn't be hit. We have the `AreUnsupportedOptions` check logged as well. I can remove this line. Thoughts?

mslukebo: I would say remove it for two reasons:
1. Debug asserts are more for the developers of the code being asserted, which wouldn't be plugin authors. They should signify an invariant of a system has been broken. Because a resolver can always return null, it is not an invariant that we don't have a null value here
2. If `AreUnsupportedOptions` logs an error if the options are null, we'll terminate processing anyway and the user will see an error anyway

anramanath: Adding a logger.warn instead and will continue since technically I suppose null processor options could be supported.

mslukebo: Hmm, what are your thoughts on documenting `IProcessorOptionsResolver.GetProcessorOptions` may not return null? This way we abort processing early if the implementation is out of spec. Processing with this as null might cause headaches in the future

mslukebo (approved):

src/Microsoft.Performance.Toolkit.Engine/Engine.cs
```diff
@@ -983,59 +983,146 @@ private RuntimeExecutionResults ProcessCore()
...
+            ProcessorOptions processorOptions, 
+            ILogger logger)
+        {
+            bool areUnsupportedOptions = false;
+
+            if (processorOptions is null)
+            {
+                logger.Warn($"Provided Processor options are null for ${processingSourceReference} (${processingSourceReference.Guid}).");
+                return true;
+            }
```
jmaxson-ms, line 1098: processorOptions should never be null, should they? if not, can we make this an assert as it's a private method?

jmaxson-ms: ah, nevermind. I missed something earlier.

src/Microsoft.Performance.Toolkit.Engine/EngineCreateInfo.cs
```diff
@@ -38,6 +39,52 @@ public EngineCreateInfo(ReadOnlyDataSourceSet dataSources)
...
+        ///     Specifies an <see cref="IProcessingOptionsResolver"/> to provide <see cref="ProcessorOptions"/> to use when processing data sources.
+        /// </summary>
+        /// <remarks>
+        ///     To set processor options for all processing sources use <see cref="WithProcessorOptions(ProcessorOptions)"/>.
+        ///     To set processor options per processing source use <seealso cref="WithProcessorOptions(IDictionary{Guid, ProcessorOptions})"/>.
+        /// </remarks>
+        /// <exception cref="ArgumentNullException">
+        ///     <paramref name="processingOptionsResolver"/> is <c>null</c>.
+        /// </exception>
+        public void WithProcessorOptions(IProcessingOptionsResolver processingOptionsResolver)
```
jmaxson-ms, line 83: >WithProcessorOptions [](http://example.com/codeflow?start=20&length=20)

Because this is "With" and not "Set", it might make more sense to return "this" to get syntax like this:

var engineCreateInfo = new EngineCreateInfo(...).With(myResolver);

jmaxson-ms: Or optionally rename it to a "Set" method.

jmaxson-ms: I don't know if this is industry standard, but it's convention in the SDK. If you do a search, you'll see that all "With" methods created/used in this project return something - often time "this".

mslukebo: +1 to return `this`. Eventually I would like customers to be able to write

```cs
var engineCreateInfo = new EngineCreateInfo()
    .WithProcessorOptions(foo)
    .WithGroupResolver(bar);
```

jmaxson-ms (approved):

mslukebo (approved):

PR #226, opened 2022-11-30
Require ISourceDataCookerFactoryRetrieval instead of ISourceDataCookerRepository in ApplicationEnvironment constructor

mslukebo:
The property this parameter gets assigned to is of type `ISourceDataCookerFactoryRetrieval `.


jmaxson-ms (approved):
:shipit:

helenkzhang (approved):

PR #252, opened 2023-01-31
Surface discoverer in AvailablePlugin

helenkzhang:
Stores which discoverer the AvailablePlugin is created from. This way it supports getting other versions and metadata directly from this discoverer without having to go back and look for the right discoverer.

### DIFF
--- a/src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core/AvailablePlugin.cs
+++ b/src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core/AvailablePlugin.cs
@@ -20,7 +20,8 @@ namespace Microsoft.Performance.Toolkit.PluginsManager.Core
             string displayName,
             string description,
             Uri packageUri,
-            Guid fetcherTypeId) 
+            Guid fetcherTypeId,
+            IPluginDiscoverer discoverer) 
         {
             this.Identity = pluginIdentity;
             this.PluginSource = pluginSource;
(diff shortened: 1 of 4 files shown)


src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core/AvailablePlugin.cs
```diff
@@ -20,14 +20,16 @@ public sealed class AvailablePlugin
             string displayName,
             string description,
             Uri packageUri,
-            Guid fetcherTypeId) 
+            Guid fetcherTypeId,
+            IPluginDiscoverer discoverer)
```
anramanath, line 24: If we have the pluginDiscoverer, instead of having the person instantiating AvailablePlugin finding the pluginSource and the URI, should we extrapolate that information here in the ctor?

Is it possible for the package URI and PluginSource to be different from the ones on the IPluginDiscoverer?

helenkzhang: Good call out. Will extrapolate `pluginSource` from the discoverer. The `packageUri` should be specified by the caller though, as it's pointing to where the plugin package should be downloaded from.

helenkzhang: See https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/253 for the alternative approach to solve this problem.

PR #253, opened 2023-02-02
Add resource guid to identify resource from an AvailablePlugin

helenkzhang:
Instead of surfacing discoverer in `AvailablePlugin` (see PR https://github.com/microsoft/microsoft-performance-toolkit-sdk/pull/252 which I've changed back to a draft).
this might be a better idea as this keeps `AvailablePlugin` being a DTO.

For context - the problem I'm trying to solve is to be able to retrieve the discoverer from an `AvailablePlugin` so that metadata can be requested after `AvailablePlugin` has been created.


src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core.NuGet/NuGetPluginDiscoverer.cs
```diff
@@ -105,7 +111,8 @@ public sealed class NuGetPluginDiscoverer : IPluginDiscoverer
                         packageMetadata.Title,
                         packageMetadata.Description,
                         this.pluginSource.Uri,
-                        Guid.Parse(PluginManagerConstants.NuGet));
+                        FetcherResourceId,
```
mslukebo, line 114: Bit confused here. Shouldn't the fetcher ID be taken from the metadata of the plugin that gets pulled?

src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core/AvailablePlugin.cs
```diff
@@ -56,8 +58,15 @@ public sealed class AvailablePlugin
         public Uri PluginPackageUri { get; }
 
         /// <summary>
-        ///     Gets the Guid which identifies the platform (NuGet, Github etc.) the plugin package is hosted on.
+        ///     Gets the Guid which identifies the unique <see cref="Transport.IPluginFetcher"/> resouce
+        ///     the plugin package should be fetched from.
         /// </summary>
-        public Guid FetcherTypeId { get; }
+        public Guid FetcherResourceId { get; }
```
mslukebo, line 64: This feels silly... you'd need to fetch the available plugin to know what this value should be. If you fetched it already, why do you need this value?

src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core.NuGet/NuGetPluginDiscoverer.cs
helenkzhang: What metadata are you referring to? The metadata from the plugin package? I don't think the metadata should know what fetcher should be used. The discoverer should provide the fetcher id.

src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core/AvailablePlugin.cs
helenkzhang: Not quite sure what you mean, but my understanding of fetching is getting the package content.  AvailablePlugin represents a discovered plugin whose content is ready to be fetched. It needs the fetcher id to know which fetcher resource it should use to fetch the package content.

src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core/AvailablePlugin.cs
```diff
@@ -2,62 +2,97 @@
...
         /// </summary>
-        public Guid FetcherTypeId { get; }
+        /// <param name="cancellationToken">
+        ///     Signals that the caller wishes to cancel the operation.
+        /// </param>
+        /// <param name="progress">
+        ///     Indicates the progress of plugin installation.
+        /// </param>
+        /// <returns></returns>
+        public Task<Stream> GetPluginPackageStream(CancellationToken cancellationToken, IProgress<int> progress)
```
mslukebo, line 93: this probably shouldn't be here... I imagine this method should be `public Task InstallAsync(CancellationToken cancellationToken, IProgress<int> progress)`

Maybe this should be an internal method?

src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core.NuGet/NuGetPluginDiscoverer.cs
```diff
@@ -64,13 +65,14 @@ public sealed class NuGetPluginDiscoverer : IPluginDiscoverer
                 {
                     var pluginId = new PluginIdentity(packageMetadata.Identity.Id, packageMetadata.Identity.Version.Version);
 
-                    var plugin = new AvailablePlugin(
+                    // TODO: #249 serialize/deserialize - fetcher resource id should be provided by nuget
```
mslukebo, line 68: also the plugin source URI

src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core/AvailablePlugin.cs
```diff
@@ -2,62 +2,97 @@
...
-        public Uri PluginPackageUri { get; }
+        /// <param name="cancellationToken">
+        ///     Signals that the caller wishes to cancel the operation.
+        /// </param>
+        /// <returns>
+        ///      A collection of available plugins.
+        /// </returns>
+        public async Task<IReadOnlyCollection<AvailablePlugin>> GetAllVersions(CancellationToken cancellationToken)
+        {
+            IReadOnlyCollection<AvailablePluginInfo> pluginInfos = await this.pluginDiscoverer.DiscoverAllVersionsOfPlugin(
```
mslukebo, line 77: There's a slight catch here... suppose we have two discoverer feeds, A and B
A has versions 1.0 and 2.0
B has versions 2.0 and 3.0

When we fetched the latest version, we returned B's version of 3.0

if we get all versions, we'll only see 2.0 and 3.0. Ideally we would get 1.0, 2.0 and 3.0

mslukebo: Would it therefore make sense to keep this method on the plugins manager? maybe passing in a PluginIdentity instead of the AvailablePlugin?

src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core/Manager/PluginsManager.cs
```diff
@@ -113,57 +113,24 @@ public void LoadAdditionalPluginResources(string directory)
...
-                        availablePlugin.Identity,
-                        cancellationToken);
-
-                    if (plugins.Any())
-                    {
-                        return plugins;
-                    }
+                    IPluginFetcher fetcher = await GetPluginFetcher(info);
+                    var AvailablePlugin = new AvailablePlugin(info, discoverer, fetcher);
+                    results.Add(AvailablePlugin);
```
mslukebo, line 120: This should probably be keeping a `Dictionary<PluginIdentity, AvailablePluginInfo> latestVersion` to return the latest versions of a particular plugin found across multiple discoverers (see my other comment about two sources and 3 versions)

src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core/AvailablePlugin.cs
mslukebo: Resolved offline

src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core.NuGet/NuGetPluginDiscoverer.cs
mslukebo: This was in reference the FetcherResourceId. I see there's the TODO

src/PluginsManager/Microsoft.Performance.Toolkit.PluginsManager.Core/Manager/PluginsManager.cs
helenkzhang: I see what you are saying. 
I think there's some confusion between discoverer and source though.  For multiple sources, definitely there can be different latest version of a same plugin and we should return all of them. For one particular source that's supported by multiple discoverers, I wonder what it means when different discoverers return different results. 

Noted that the method on L102 gets available plugins from one particular source. The method on L88 gets available plugins from all sources.

mslukebo: Yeah, in my example I'm imagining 1 source == 1 discoverer. I think if we ever are in a situation where 1 source has multiple discoverers, we shouldn't even allow adding that source

helenkzhang: @mslukebo  After thinking about this for a bit - we should still allow multiple discoverers to attempt to discover from one source/feed because there isn't so much `IPluginDiscovererProvider.IsSupportedAsync` can check from a URI. I have updated the logic for handling duplicated within one feed and across multiple feed and added some explanatory comments inline. Can you take another look?

mslukebo (approved):

PR #258, opened 2023-02-06
Delay creation of source processing session.

jmaxson-ms:
'this' is passed into CreateSourceSession, but this was being called from the constructor of an abstract class. If the source processing session factory, or the implementation of that interface, relies on something in the class to be initialized, then this would break.

Making the creation of the source processing session lazy means that we don't introduce any breaking changes while allowing more flexibility in the generation/implementation of ISourceProcessingSession.

### DIFF
--- a/src/Microsoft.Performance.SDK/Processing/CustomDataProcessorWithSourceParser.cs
+++ b/src/Microsoft.Performance.SDK/Processing/CustomDataProcessorWithSourceParser.cs
@@ -45,8 +45,9 @@ namespace Microsoft.Performance.SDK.Processing
 
             this.SourceParser = sourceParser;
 
+            var sourceSessionFactory = applicationEnvironment.SourceSessionFactory;
             this.sourceProcessingSession = new Lazy<ISourceProcessingSession<T, TContext, TKey>>(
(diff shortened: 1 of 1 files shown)


mslukebo (approved):

src/Microsoft.Performance.SDK/Processing/CustomDataProcessorWithSourceParser.cs
```diff
@@ -42,7 +44,10 @@ public abstract class CustomDataProcessorWithSourceParser<T, TContext, TKey>
             Guard.NotNull(sourceParser, nameof(sourceParser));
 
             this.SourceParser = sourceParser;
-            this.SourceProcessingSession = this.ApplicationEnvironment.SourceSessionFactory.CreateSourceSession(this);
+
+            this.sourceProcessingSession = new Lazy<ISourceProcessingSession<T, TContext, TKey>>(
+                () => this.ApplicationEnvironment.SourceSessionFactory.CreateSourceSession(this),
```
trgibeau, line 49: Lets store the SourceSessionFactory that we can reference it directly

trgibeau (approved):

mslukebo (approved):

PR #263, opened 2023-02-10
Add Plugins System Core and Runtime

helenkzhang:
Introduced **Plugins System** which can be used to discover and install performance analysis plugins. The two libraries added in the PR are:
- `Microsoft.Performance.Toolkit.Plugins.Core` - contracts and core functionalities for implementing plugins discovering and fetching. 
- `Microsoft.Performance.Toolkit.Plugins.Runtime` - contracts and a default implementation of `PluginsSystem` which can be used by plugins consumers to manage plugins.


src/Microsoft.Performance.SDK.Runtime/PluginsManagerResourceReference.cs
```diff
@@ -0,0 +1,142 @@
+﻿// Copyright (c) Microsoft Corporation.
```
trgibeau, line 1: Not sure if its GitHub, just noticed the weird chars at the top here

src/Microsoft.Performance.SDK/HashUtils.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation.
```
trgibeau, line 1: Not sure if its GitHub, just noticed the weird chars at the top here

src/Microsoft.Performance.SDK/HashUtils.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// </param>
+        /// <returns>
+        ///     The combined MD5 hash of all files including the file paths.
+        /// </returns>
+        public static async Task<string> GetDirectoryHash(string directory)
+        {
+            Guard.NotNull(directory, nameof(directory));
+            Guard.IsTrue(Directory.Exists(directory), $"Directory {directory} doesn't exist.");
+
+            using (var md5 = MD5.Create())
```
trgibeau, line 31: Do we want to use the SHA256 class instead of MD5?

Due to collision problems with MD5/SHA1, Microsoft recommends SHA256 or SHA512. Consider using the [SHA256](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha256?view=net-7.0) class or the [SHA512](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha512?view=net-7.0) class instead of the [MD5](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.md5?view=net-7.0) class. Use [MD5](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.md5?view=net-7.0) only for compatibility with legacy applications and data.

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Extensibility/PluginsManagerResourceReflector.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <inheritdoc />
+        public void ProcessType(Type type, string sourceName)
+        {
+            if (PluginsManagerResourceReference.TryCreateReference(type, out PluginsManagerResourceReference reference))
+            {
+                try
+                {
+                    this.loadedResources.Add(type, reference);
+                }
+                catch { }
```
trgibeau, line 43: should we catch and log/return an error?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsManager.cs
```diff
@@ -0,0 +1,624 @@
...
+            try
+            {
+                await task.ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+                this.logger.Info($"The request to get lastest available plugins is cancelled.");
+                throw;
+            }
+            catch
```
trgibeau, line 156: Should we catch this exception too?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsManager.cs
```diff
@@ -0,0 +1,624 @@
...
+            try
+            {
+                await task.ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+                this.logger.Info($"The request to get available plugins from {pluginSource} is cancelled.");
+                throw;
+            }
+            catch
```
trgibeau, line 241: should we catch this exception too?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsManager.cs
```diff
@@ -0,0 +1,624 @@
...
+            try
+            {
+                await task.ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+                this.logger.Info($"The request to get all versions of plugin {pluginIdentity} is cancelled.");
+                throw;
+            }
+            catch
```
trgibeau, line 302: same here

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsManager.cs
```diff
@@ -0,0 +1,624 @@
...
+            var task = Task.WhenAll(tasks);
+            try
+            {
+                await task.ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+                this.logger.Info($"The request to get all versions of plugin {pluginIdentity} from {source} is cancelled.");
+            }
+            catch
```
trgibeau, line 348: same here

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsManager.cs
helenkzhang: Since this is running in a Task.WhenAll, this catch here won't be able to handle exceptions from each individual task. They are instead handled in the for loop below. Same for other places that use Task.WhenAll.

src/Microsoft.Performance.SDK/HashUtils.cs
helenkzhang: Updated

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Extensibility/PluginsManagerResourceReflector.cs
helenkzhang: The reflector is disabled right now. Added a TODO here so that we handle errors once it's re-enabled.

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsManager.cs
helenkzhang: Added a comment

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/Discovery/IPluginDiscoverer.cs
```diff
@@ -0,0 +1,68 @@
...
+        ///     Discovers the latest version of all plugins.
+        /// </summary>
+        /// <param name="cancellationToken">
+        ///     Signals that the caller wishes to cancel the operation.
+        /// </param>
+        /// <returns>
+        ///     The <see cref="AvailablePluginInfo"/> of all discovered plugin grouped by plugin identifiers.
+        /// </returns>
+        /// TODO: 243 Add search
+        Task<IReadOnlyDictionary<string, AvailablePluginInfo>> DiscoverPluginsLatestAsync(CancellationToken cancellationToken);
```
mslukebo, line 27: Why is this a dictionary? What are the keys supposed to be again?

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/AvailablePlugin.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            this.AvailablePluginInfo = pluginInfo;
+            this.pluginDiscoverer = discoverer;
+            this.pluginFetcher = fetcher;
+        }
+
+        /// <summary>
+        ///     Gets the <see cref="AvailablePluginInfo"/> associated with this plugin.
+        /// </summary>
+        public AvailablePluginInfo AvailablePluginInfo { get; }
```
mslukebo, line 48: nit: I would personally name this property just "Info." Accessing this as `AvailablePlugin.AvailablePluginInfo` is a bit redundant IMO

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/AvailablePlugin.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <returns>
+        ///     The metadata of a plugin.
+        /// </returns>
+        public Task<PluginMetadata> GetPluginMetadata(CancellationToken cancellationToken)
+        {
+            return this.pluginDiscoverer.GetPluginMetadataAsync(this.AvailablePluginInfo.Identity, cancellationToken);
+        }
+
+        /// <summary>
+        ///     Gets stream to this plugin that can be installed via IPluginsInstaller.Install()
```
mslukebo, line 65: nit: when referencing other types in a docstring, use `<see cref="" />` tags

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/Discovery/IPluginDiscoverer.cs
helenkzhang: The keys are `PluginIdentity.Id`s. This is to make sure each plugin id gets only "one latest" `AvailablePluginInfo`. I'll update the docstring so that it's not confusing.

mslukebo: I feel like we should just ditch the dictionary altogether since I'm not sure what it adds
1. Clients will likely just completely ignore the keys
2. The method spec doesn't actually enforce anything. Implementations can return a dictionary with bogus values for keys, and different versions of the same plugin as values. Whoever actually consumes this method should be responsible for filtering out duplicate results

mslukebo: Case in point: `IPluginsDiscoveryOrchestrator` doesn't return a dictionary

IMO, this return signature just makes correct implementations awkward. Most likely an implementation will just be doing something like
```cs
return latestVersions.ToDictionary(p => p.Identity.Id, p => p);
```

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/ErrorCodes.cs
```diff
@@ -0,0 +1,187 @@
...
+
+namespace Microsoft.Performance.Toolkit.Plugins.Runtime
+{
+    /// <summary>
+    ///     This class represents all of the error codes that can be
+    ///     emitted by the plugins system runtime. This class acts as a "type-safe"
+    ///     enumeration. All new explicit error conditions should be
+    ///     added to this class.
+    /// </summary>
+    public sealed class ErrorCodes
```
mslukebo, line 17: Can we make the other ErrorCodes class partial, the make this another part of the class? Instead of duplicating a bunch of logic

mslukebo: I.e. make both classes `public sealed partial class`

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/InstalledPlugin.cs
```diff
@@ -0,0 +1,40 @@
...
+            PluginMetadata pluginMetadata)
+        {
+            this.PluginInfo = installedPluginInfo;
+            this.PluginMetadata = pluginMetadata;
+        }
+
+        /// <summary>
+        ///     Gets the information of the installed plugin.
+        /// </summary>
+        public InstalledPluginInfo PluginInfo { get; }
```
mslukebo, line 33: nit: I would also change these properties to "Info" and "Metadata"

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Package/PackageConstants.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License.
+
+namespace Microsoft.Performance.Toolkit.Plugins.Runtime.Package
+{
+    /// <summary>
+    ///     Constants used by the default plugin package.
+    /// </summary>
+    public static class PackageConstants
+    {
+        public const string PluginMetadataFileName = "pluginspec.json";
```
mslukebo, line 11: is this file still needed?

documentation/Plugins-System-Preview-Only/Plugins-Glossary-Preview-Only.md
```diff
@@ -0,0 +1,29 @@
...
+
+**Processing Source** - An entry point for a plugin. A processing source declares a human-readable name and description, defines the data sources it supports, creates the data processor that will process instances of supported data sources and advertises the tables that may be built in response to data sources being processed
+
+**Plugin Package** - A zip file with a custom extension (e.g. *.plugin*) containing the content of a plugin along with a metadata file that describes the plugin(s).  A plugin package may contain only one plugin. In other words, whatever is packaged into a plugin package by the the plugin author will be called **a** plugin in our new ecosystem. For v1, a plugin package will be self-contained in the way that it doesn't have dependencies on any other non-plugin packages. For example, a plugin must include Newtonsoft.Json if it requires it for JSON parsing.
```
mslukebo, line 6: If a package may contain only one plugin, remove the (s) from the preceding sentence

documentation/Plugins-System-Preview-Only/Plugins-Glossary-Preview-Only.md
```diff
@@ -0,0 +1,29 @@
...
+
+**Plugin Consumer** - A client software that consumes a plugin package to get access to plugins and their processing sources. E.g., wpa.exe, the Engine.
+- **Plugin Loader** - Used by the plugin consumer to load a set of plugins supplied by the plugin packages.
+
+**Plugins System** - Used by the plugin consumer to manage the discovering, installing, uninstalling and updating of plugins.
+
+#### Different Forms of Plugin
+ - **Available Plugin**  - A plugin package discovered from a plugin source that is available for a plugin consumer to install.
+ - **Installed Plugin** - A plugin that a user has opted to install. Installed plugins reside on the user's machine and can be loaded by a plugin consumer.
+  - **Loaded Plugin** - A plugin that has been previously installed from a plugin package and has been loaded by the plugin loader for trace processing within an instance of a plugin consumer being opened.
```
mslukebo, line 18: Current impl does not define this. Should we remove?

documentation/Plugins-System-Preview-Only/Plugins-Glossary-Preview-Only.md
```diff
@@ -0,0 +1,29 @@
...
+
+**Plugin Source** - A source (e.g. a NuGet feed, a REST API, a text file) where plugins can be discovered from. The source provides information about how a plugin package can be transported to a plugin consumer.
+
+**Plugin Installation** - A plugin installation involves two parts:
+1. Taking a plugin package and processing its contents - extracting contents to some folder.
+2. Recording installation to a *registry* - noting which version of which plugin was installed to which location.
+Installation does not require a feed. A plugin can be installed from a local plugin package.
+
+**Plugin Registry** - A local registry has two components: a registry file that records information of installed plugins and an ephemeral lock file that indicates whether a process is currently interacting with the registry. Local registry is on a per-consumer basis. (e.g. WPA may have its own registry that is different from some other consumer using the engine)
+- **Plugin Package Cache** - Simply a directory containing plugin packages that may currently be installed.
```
mslukebo, line 28: Do we do this?

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/PluginSourceRepository.cs
```diff
@@ -0,0 +1,142 @@
...
+using Microsoft.Performance.Toolkit.Plugins.Core.Discovery;
+using Microsoft.Performance.Toolkit.Plugins.Runtime.Common;
+
+namespace Microsoft.Performance.Toolkit.Plugins.Runtime
+{
+    /// <summary>
+    ///    Represents a repository of <see cref="PluginSource"/>s.
+    /// </summary>
+    public sealed class PluginSourceRepository
+        : IRepository<PluginSource>
```
mslukebo, line 20: this could probably be abstracted into some type of generic abstract class

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Package/PackageConstants.cs
helenkzhang: Yes

documentation/Plugins-System-Preview-Only/Plugins-Glossary-Preview-Only.md
helenkzhang: No, will remove

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/ErrorCodes.cs
helenkzhang: Looks like each part of a partial class should be in the same namespace and the same assembly.

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/Discovery/IPluginDiscoverer.cs
helenkzhang: Updated

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/PluginSourceRepository.cs
helenkzhang: Updated

mslukebo (approved):

mslukebo (approved):

anramanath (approved):

anramanath (approved):

mslukebo (approved):

PR #268, opened 2023-02-23
Plugins Manager Error Handling and Logging

helenkzhang:
Changes:
- Remove public setters in metadata classes
- Support setting loggers in resource types
- Add loggers in runtime types
- Add custom plugins manager exceptions
- Log and throw for install/uninstall operations
- Catch and log individual discovery exceptions
- Refactor plugin registry class
- Discovery error reporting through events


src/Microsoft.Performance.SDK/Processing/ProcessingSourceInfo.cs
```diff
@@ -41,6 +68,29 @@ public sealed class ProcessingSourceInfo
...
+            this.Name = name;
+            this.Address = address;
+            this.EmailAddresses = emailAddresses;
+            this.PhoneNumbers = phoneNumbers;
+        }
+
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="ContactInfo"/> class.
+        /// </summary>
+        public ContactInfo()
```
mslukebo, line 90: What's the reason for these empty ctors?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Core.NuGet/NuGetPluginDiscoverer.cs
```diff
@@ -87,7 +86,7 @@ public sealed class NuGetPluginDiscoverer : IPluginDiscoverer
             PackageSearchResource searchResource = await this.repository.GetResourceAsync<PackageSearchResource>(cancellationToken);
 
             var result = new Dictionary<string, AvailablePluginInfo>();
-            IEnumerable<IPackageSearchMetadata> packages = await searchResource.SearchAsync(null, new SearchFilter(false), 0, PageCount, this.logger, cancellationToken);
+            IEnumerable<IPackageSearchMetadata> packages = await searchResource.SearchAsync(null, new SearchFilter(false), 0, PageCount, null, cancellationToken);
```
mslukebo, line 89: Instead of passing null, I would either
1. Remove the parameter entirely
2. Pass some kind of "null logger"
3. Pass `Logger.Create(typeof(thing))` (preferable if the parameter is needed)

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Core/Transport/IPluginFetcher.cs
```diff
@@ -13,26 +14,33 @@ namespace Microsoft.Performance.Toolkit.Plugins.Core.Transport
...
         /// </param>
+        /// <param name="logger">
+        ///     Used to log information.
+        /// </param>
         /// <returns>
         ///     <c>true</c> if <paramref name="pluginInfo"/> is supported by this fetcher. <c>false</c>
         ///     otherwise.
         /// </returns>
-        Task<bool> IsSupportedAsync(AvailablePluginInfo pluginInfo);
+        Task<bool> IsSupportedAsync(AvailablePluginInfo pluginInfo, ILogger logger);
```
mslukebo, line 33: I am not a fan of the discoverer having a `SetLogger` method, but the fetcher getting them passed through method parameters

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Result/InstalledPluginsResults.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.Performance.Toolkit.Plugins.Runtime.Result
+{
+    public class InstalledPluginsResults
+    {
+        public InstalledPluginsResults(
+                        IReadOnlyCollection<InstalledPlugin> validPlugins,
+            IReadOnlyCollection<InstalledPluginInfo> invalidPluginsInfo)
```
mslukebo, line 12: How does this work exactly? If there's some problem fetching info for a plugin, how do we have enough information to create an `InstalledPluginInfo` for it?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/PluginsManagerException.cs
```diff
@@ -0,0 +1,281 @@
+﻿using System;
+using Microsoft.Performance.Toolkit.Plugins.Core;
+
+namespace Microsoft.Performance.Toolkit.Plugins.Runtime
```
mslukebo, line 4: I think some of these classes should be `abstract` (e.g. `PluginsManagerException`)

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/PluginsManagerException.cs
```diff
@@ -0,0 +1,281 @@
...
+            : base(message, innerException)
+        {
+        }
+    }
+
+    //
+    // PluginRegistry Exception Types
+    //
+    
+    public class PluginRegistryException
```
mslukebo, line 28: Can you create a new sub-folder and put all these classes in their own files?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/IPluginsManager.cs
```diff
@@ -112,9 +117,12 @@ public interface IPluginsManager
         ///     Signals that the caller wishes to cancel the operation.
         /// </param>
         /// <returns>
-        ///     A collection of installed plugins.
+        ///     A <see cref="InstalledPluginsResults"/> instance containing all valid and invalid installed plugins.
         /// </returns>
-        Task<IReadOnlyCollection<InstalledPlugin>> GetInstalledPluginsAsync(CancellationToken cancellationToken);
+        /// <exception cref="OperationCanceledException">
+        ///     The operation was canceled.
+        /// </exception>
```
mslukebo, line 124: I assume this can throw other exceptions, like if the registry is corrupt and what not?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/IPluginsManager.cs
```diff
@@ -142,21 +165,29 @@ public interface IPluginsManager
...
         /// <returns>
-        ///     <c>true</c> if the plugin has been successfully installed. <c>false</c> otherwise.
+        ///     The <see cref="InstalledPluginInfo"/> if plugin is successfully installed.. <c>null</c> otherwise.
         /// </returns>
-        Task<bool> InstallLocalPluginAsync(
+        /// <exception cref="ArgumentNullException">
+        ///     Throws when <paramref name="pluginPackagePath"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="PluginRegistryException">
+        ///     Throws when there is an error interacting with plugin registry.
```
mslukebo, line 181: I would prefer if these errors were more fine-grained. Some exceptions can be recovered from, others cannot.

For example, if the issue is that the plugin registry is corrupt, we can recover by simply deleting the registry and starting over. But, if the issue is that the registry cannot be accessed, then we have a bigger problem since we need to change its location

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/IPluginsManager.cs
```diff
@@ -129,9 +137,24 @@ public interface IPluginsManager
...
+        /// </exception>
+        /// <exception cref="PluginPackageException">
+        ///     Throws when there is an error interacting with plugin package.
+        /// </exception>
+        /// <exception cref="OperationCanceledException">
+        ///     Throws when the operation was canceled.
+        /// </exception>
+        Task<InstalledPluginInfo> InstallAvailablePluginAsync(
```
mslukebo, line 157: Mind renaming this to `TryInstallAvailablePluginAsync`?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/IPluginsManager.cs
```diff
@@ -142,21 +165,29 @@ public interface IPluginsManager
...
+        /// <exception cref="PluginRegistryException">
+        ///     Throws when there is an error interacting with plugin registry.
+        /// </exception>
+        /// <exception cref="PluginPackageException">
+        ///     Throws when there is an error interacting with plugin package.
+        /// </exception>
+        /// <exception cref="OperationCanceledException">
+        ///     Throws when the operation was canceled.
+        /// </exception>
+        Task<InstalledPluginInfo> InstallLocalPluginAsync(
```
mslukebo, line 189: Same here, `TryInstallLocalPluginAsync`

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Registry/PluginRegistry.cs
```diff
@@ -198,11 +347,26 @@ private async Task<List<InstalledPluginInfo>> ReadInstalledPlugins()
...
+            try
+            {
+                using (var fileStream = new FileStream(this.registryFilePath, FileMode.Open, FileAccess.Read, FileShare.Read))
+                {
+                    return await JsonSerializer.DeserializeAsync<List<InstalledPluginInfo>>(
+                        fileStream,
+                        SerializationConfig.PluginsManagerSerializerDefaultOptions);
+                }
+            }
+            catch (Exception e)
```
mslukebo, line 360: I don't think this should be a catch-all. There's probably only a subset of exceptions we expect to happen here

src/Microsoft.Performance.SDK/Processing/ProcessingSourceInfo.cs
helenkzhang: These are currently used by `ProcessingSource` in `GetAboutInfo()`. Would be a breaking change to remove these.

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Core/Transport/IPluginFetcher.cs
helenkzhang: Trying to keep fetcher thread safe by not adding properties to it (assuming logger is not thread safe?) , as fetcher would be a singleton for plugins manager and can be accessed by multiple threads.  But now that I think about it, I should do the same for `IPluginDiscovererProvider` as all the resources are on the same boat. 

I'm leaning towards `SetLogger` for all of them and use locks in the callers. Any thoughts?

mslukebo: Yes, `SetLogger` makes more sense. I would assume it's up to the clients to make sure loggers aren't being concurrently set with calls to fetch/discover things.

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Result/InstalledPluginsResults.cs
helenkzhang: This is used only by `GetInstalledPluginsAsync` which does three things: 1. Read all records from the plugin registry  2. For each record, read their metadata from where they're installed. 3. For each record, create a `InstalledPlugin` instance and return all the created `InstalledPlugin`s. 

This method will throw `PluginRegistryException`s  but will swallow exceptions thrown by individual plugin records when reading the metadata.  For those that fail to read metadata, they are added to `InvalidPluginsInfo` in this result class (they will always have `InstalledPluginInfo`) so that the client is aware of these plugins.

mslukebo: Got it. For some reason, I assumed the metadata was also on InstalledPluginInfo

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Registry/PluginRegistry.cs
helenkzhang: The catch is basically to wrap the exception happens during the read and write of plugin registry file in a custom type and re-throw.  Would you prefer system exceptions like `UnauthorizedAccessException` to not be wrapped? Maybe only expect JsonException and IOException?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Core/Transport/IPluginFetcher.cs
helenkzhang: I've updated to use `SetLogger()`. However, the thing I was worried about is not only the Set itself but also the thread safety of other public methods on `IPluginFetcher` and `IPluginDiscoverer` because they are shared resources and the logger (the default one uses a `TextWriter` under the hood which is not thread safe). Wonder if I would need to ensure thread safety in the callers of all of these methods.

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Core/Extensibility/IPluginManagerResource.cs
```diff
@@ -1,13 +1,30 @@
...
     public interface IPluginManagerResource
     {
+        /// <summary>
+        ///     Provides the <see cref="IPluginManagerResource"/> an application-appropriate logging mechanism.
+        /// </summary>
+        /// <param name="logger">
+        ///     Used to log information.
+        /// </param>
+        /// <remarks>
+        ///     Ensure that this method called and called only once before the resource is used.
```
mslukebo, line 25: nit: "is called"

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Events/PluginSourceErrorEventArgs.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// </summary>
+    public abstract class PluginSourceErrorEventArgs
+        : EventArgs
+    {
+        protected PluginSourceErrorEventArgs(PluginSource pluginSource)
+            : this(pluginSource, $"Error occurred when interacting with plugin source: {pluginSource.Uri}")
+        {
+        }
+
+        protected PluginSourceErrorEventArgs(PluginSource pluginSource, string errorMessage)
```
mslukebo, line 20: nit: may want to have an optional `ExceptionDispatchInfo` on here

mslukebo: Hmm, I see the other class. I would prefer just having one sealed class that has an `Exception` property that may be null

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsManager.cs
```diff
@@ -582,21 +589,37 @@ private async Task<IPluginFetcher> TryGetPluginFetcher(AvailablePluginInfo avail
...
+
+        private void HandleResourceNotFoundError(PluginSource source, string errorMsg)
+        {
+            var errorInfo = new ErrorInfo(ErrorCodes.PLUGINS_MANAGER_PluginManagerResourceNotFound, errorMsg);
+            PluginSourceErrorOccured?.Invoke(this, new PluginSourceErrorEventArgs(source, errorInfo));
+
+            this.logger.Error(errorMsg);
+        }
+
+        private void HandlePluginSourceException(PluginSource source, string errorMsg, Exception exception)
```
anramanath, line 617: I haven't looked at the whole PR yet so I might lack context. I see we added some Exception Error codes too. 

Do you think it might be useful to create a new `PluginException` which extends the Exception class similar to `EngineException` so we can reuse this exception in the future?

src/Microsoft.Performance.SDK.Runtime/ErrorCodes.cs
```diff
@@ -212,6 +212,26 @@ public sealed class ErrorCodes
...
+        //
+        // Plugins Manager Errors
+        //
+
+        /// <summary>
+        ///    An error occurred when interacting with a plugin source.
+        /// </summary>
+        public static ErrorCodes PLUGINS_MANAGER_PluginSourceException = new ErrorCodes(
+            50001,
+            "PLUGIN_MANAGER_PluginSourceException",
```
anramanath, line 224: Should we name this `PLUGINS_MANAGER_PluginSourceError` to be more accurate?

anramanath (approved):

PR #272, opened 2023-03-08
Minor refactor of plugins manager

helenkzhang:
- Abstract out ISerializer
- Abstract out IRepository from IPluginRegistry
- Abstract out IPluginInstaller
- PluginPackageInstaller requires ISerializer and IPluginRegistry as dependencies
- PluginsManager requires IPluginInstaller as a dependency


src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/IRepository`1.cs
```diff
@@ -0,0 +1,161 @@
...
+        /// <exception cref="Exceptions.RepositoryCorruptedException">
+        ///     Throws when the repository is in an invalid state.
+        /// </exception>
+        /// <exception cref="Exceptions.RepositoryDataAccessException">
+        ///     Throws when the repository cannot be read or written.
+        /// </exception>
+        /// <exception cref="ArgumentNullException">
+        ///     Throws when <paramref name="Id"/> is null.
+        /// </exception>
+        Task<TEntity> TryGetByIdAsync(string Id, CancellationToken cancellationToken);
```
mslukebo, line 57: should the ID be a type parameter?

helenkzhang: Updated

PR #275, opened 2023-03-15
Refactor plugins manager so that it doesn't have implicit file system dependencies 

helenkzhang:


src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/IPluginsManager.cs
```diff
@@ -209,7 +217,7 @@ public interface IPluginsManager
         ///     Throws when the operation was canceled.
         /// </exception>
         Task<InstalledPluginInfo> TryInstallLocalPluginAsync(
-           string pluginPackagePath,
```
mslukebo, line 212: This seems confusing. Why can't this just take in a `Stream`? Considering the plugin fetchers all return streams, it should be possible to install a local `Stream`

PR #280, opened 2023-03-16
Refactor IPluginsManager to a composition-based PluginsSystem 

helenkzhang:
Cleaning up Plugins Manager will be in another PR.
Changes in this PR: 

1. `DiscoverySourcesManager`  -> 
    - `PluginsDiscoverer` which implements `IPluginsDisoverer`
    - `PluginSourceRepository` which implements `IRepository<PluginsSource>`
2. Add synchronization in `PluginsDiscoverer`
4. Remove installPath from `InstalledPluginInfo` and have installation directory passed as a constructor parameter to `FileBackedPluginsInstaller`.
5. Remove registry root from `IPluginRegistry`
6. Make `AvailablePlugin` constructor internal and make `GetPluginPackageStream` public


src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Common/IReadonlyRepository`1.cs
```diff
@@ -0,0 +1,28 @@
...
+    ///     Represents a repository for storing a collection of entities of type <typeparamref name="TEntity"/>.
+    ///     Supports read-only operations and notifications when the repository is modified.
+    /// </summary>
+    /// <typeparam name="TEntity">
+    ///     The type of the resources stored in this repository.
+    /// </typeparam>
+    public interface IReadonlyRepository<TEntity>
+    {
+        /// <summary>
+        ///     Gets all plugins manager resources contained in this repository.
```
mslukebo, line 19: nit: update docstring

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Common/IRepository`2.cs
```diff
@@ -6,7 +6,7 @@
 using System.Threading.Tasks;
 using System.Threading;
```
mslukebo, line 7: I would consider renaming this to something more appropriate now like `IKeyedRepository`

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Discovery/IPluginsDiscoveryManager.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    /// <summary>
+    ///     Contains logic for discovering plugins.
+    /// </summary>
+    public interface IPluginsDiscoveryManager
+    {
+        /// <summary>
+        ///     Gets all plugin sources that this plugin manager is configured to discover plugins from.
+        /// </summary>
+        IEnumerable<PluginSource> PluginSources { get; }
```
mslukebo, line 22: I would remove this property and make this interface extend `IRepository<PluginSource>`

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/PluginSourceRepo.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Linq;
+using Microsoft.Performance.Toolkit.Plugins.Core.Discovery;
+using Microsoft.Performance.Toolkit.Plugins.Runtime.Common;
+
+namespace Microsoft.Performance.Toolkit.Plugins.Runtime
+{
+    public class PluginSourceRepo
+        : IReadonlyRepository<PluginSource>
+    {
+        private HashSet<PluginSource> currentSources;
```
mslukebo, line 12: Might want to have a RW lock around getting/setting this this

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/IPluginsManager.cs
```diff
@@ -16,108 +13,12 @@ namespace Microsoft.Performance.Toolkit.Plugins.Runtime.Manager
...
-            PluginSource pluginSource,
-            PluginIdentity pluginIdentity,
-            CancellationToken cancellationToken);
-
-        /// <summary>
-        ///    Raised when an error occurs interacting with a paticular <see cref="PluginSource"/>.
-        ///    Subsribe to this event to handle errors related to a particular <see cref="PluginSource"/>.
-        /// </summary>
-        event EventHandler<PluginSourceErrorEventArgs> PluginSourceErrorOccured;
+        Task<InstalledPluginInfo> TryInstallAvailablePluginAsync(
```
mslukebo, line 19: Why is this method not on `IPluginsDiscoveryManager`?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Common/IRepositoryRO`1.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Collections.Specialized;
+
+namespace Microsoft.Performance.Toolkit.Plugins.Runtime.Common
+{
+    /// <summary>
+    ///     Represents a repository for storing a collection of entities of type <typeparamref name="TEntity"/>.
+    ///     Supports read-only operations and notifications when the repository is modified.
+    /// </summary>
+    /// <typeparam name="TEntity"></typeparam>
+    public interface IRepositoryRO<TEntity>
```
mslukebo, line 14: I'm realizing this interface is essentially the same thing as `IReadOnlyObservableCollection`. Do we want to just use that, or keep this?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Common/IRepository`1.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace Microsoft.Performance.Toolkit.Plugins.Runtime.Common
+{
+    /// <summary>
+    ///     Represents a repository for storing a collection of entities of type <typeparamref name="TEntity"/>.
+    ///     This repository supports adding and removing items.
+    /// </summary>
+    /// <typeparam name="TEntity">
+    ///     The type of the resources stored in this repository.
+    /// </typeparam>
+    public interface IRepository<TEntity>
```
mslukebo, line 15: Same comment here (but `IObservableCollection`

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsSystem.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// <summary>
+        ///     Creates an instance of the <see cref="PluginsSystem"/>.
+        /// </summary>
+        /// <param name="installer">
+        ///     The installer to use.
+        /// </param>
+        /// <param name="discoverer">
+        ///     The discoverer to use.
+        /// </param>
+        private PluginsSystem(
```
mslukebo, line 28: Wonder if we should leave this as public, but in docstring call out the static methods below

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsSystem.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <summary>
+        ///     Gets the discoverer.
+        /// </summary>
+        public IPluginsDiscoverer Discoverer { get; }
+
+        /// <summary>
+        ///     Gets the repository of plugin fetchers.
+        /// </summary>
+        public IRepositoryRO<IPluginFetcher> FetchersRepository { get; }
```
mslukebo, line 102: Do we need this (and the one below)? These seem like an implementation detail

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/PluginSourceRepository.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Linq;
+using Microsoft.Performance.Toolkit.Plugins.Core.Discovery;
+using Microsoft.Performance.Toolkit.Plugins.Runtime.Common;
+
+namespace Microsoft.Performance.Toolkit.Plugins.Runtime
+{
+    /// <summary>
+    ///    Represents a repository of <see cref="PluginSource"/>s.
+    /// </summary>
+    public sealed class PluginSourceRepository
```
mslukebo, line 15: If we change the interface to observable collection, we could ditch this class and just use `ObservableCollection<PluginSource>`

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsSystem.cs
helenkzhang: Actually I just realized this won't work. Because we are creating these repositories in the static method hot path and only exposing a readonly interface, there's no way for the client to pass any dicoverers and fectchers down to the plugins system.   

I think we should either use another interface (supports adding new resource and interacts with resource loader) so that resources can be added later by the client, or have the client initialize these repositories. What do you think?

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsSystem.cs
helenkzhang: Yep. Changed in the 2nd PR

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Common/IRepositoryRO`1.cs
helenkzhang: Looks like  .NET doesn't have the class but no interface for `ObservableCollection` .  WPA do have them under the `Microsoft.VisualStudio.PlatformUI.Shell`

src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Manager/PluginsSystem.cs
mslukebo: Yeah, I think we may want to extract out and expose some type an `IPluginSystemResourceLoader` interface for consumers to "load" instances of discoverers/fetchers. The loader would be passed down to anyone who needs the resources as an `IRepositoryRO<T>`. For example, we may want to have:

```cs
public interface IPluginSystemResourcesLoader
{
    void Add(IPluginDiscoverer discoverer);

    void Add(IPluginFetcher fetcher);
}

public class PluginSystemResourcesLoader
    : IPluginSystemResourcesLoader,
      IRepositoryRO<IPluginDiscoverer>,
      IRepositoryRO<IPluginFetcher>
{
    ...
}
```

helenkzhang: We already have a mechanism  [IPluginsManagerResourceLoader](https://github.com/microsoft/microsoft-performance-toolkit-sdk/blob/feature/pluginManager/src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Extensibility/IPluginsManagerResourceLoader.cs) for loading resources using a subscribe/unsubscribe model similar to what the SDK runtime does today:

The [PluginsManagerResourceRepository](https://github.com/microsoft/microsoft-performance-toolkit-sdk/blob/feature/pluginManager/src/PluginsManager/Microsoft.Performance.Toolkit.Plugins.Runtime/Extensibility/PluginsManagerResourceRepository.cs)  is both an `IRepositoryRO` and an `IPluginsManagerResourcesConsumer` which can be registered to the loader to receive new resources. We can create a new interface that inherits both of them and use it as the type we expose in the `PluginsSystem` class.

helenkzhang: @mslukebo Pushed an update addressing this.

mslukebo (approved):

PR #285, opened 2023-03-27
Update version files so that PluginsSystem will be on its own version

helenkzhang:
- Exclude PluginSystem in pathFilters of the main version.json
- Add a new version.json for PluginSystem

### DIFF
--- a/src/PluginsSystem/version.json
+++ b/src/PluginsSystem/version.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "1.1-preview",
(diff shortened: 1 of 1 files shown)


src/PluginsSystem/version.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
+  "version": "1.1-preview",
```
mslukebo, line 3: Should we start these off at 0.1-preview? Don't really see the need to rush to 1.0 :)

helenkzhang: @trgibeau What do you think about the version number?

mslukebo (approved):

PR #287, opened 2023-03-30
Move version relevant files to src and exclude .sln in version path filters

helenkzhang:


src/version.json
```diff
@@ -2,12 +2,17 @@
...
+    "precision": "revision"
   },
   "publicReleaseRefSpec": [
     "^refs/heads/main$",
     "^refs/heads/release/v?\\d+(?:\\.\\d+)?$"
   ],
+  "pathFilters": [
+    ".",
+    ":!SDK.sln",
+    ":!PluginsSystem"
```
helenkzhang, line 14: We exclude `PluginsSystem` now so that version.json won't need to change when the plugins stuff gets checked in. Otherwise, SDK version will be bumped.

mslukebo (approved):

anramanath (approved):

PR #288, opened 2023-04-07
Refactor plugin installer

helenkzhang:


src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/Checksum/SHA256DirectoryChecksumCalculator.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// <summary>
+    ///     Represents a checksum calculator that uses SHA256.
+    /// </summary>
+    public sealed class SHA256DirectoryChecksumCalculator
+        : IDirectoryChecksumCalculator
+    {
+        /// <inheritdoc />
+        public async Task<string> GetDirectoryChecksumAsync(string directory)
+        {
+            return await HashUtils.GetDirectoryHashAsync(directory);
```
mslukebo, line 18: You might want to move the impl here so it's obvious that it MUST be a SHA256 (given the class name)

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/PluginsSystem.cs
```diff
@@ -84,13 +92,15 @@ public sealed class PluginsSystem
         /// </returns>
         public static PluginsSystem CreateFileBasedPluginsSystem(
             string root,
+            Func<string, Task<bool>> loadPluginFromDirectory,
```
mslukebo, line 95: we might want to consider, in the future, just making this an interface `IPluginDirectoryLoader`

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/Storage/IPluginsStorageDirectory.cs
```diff
@@ -0,0 +1,56 @@
...
+    public interface IPluginsStorageDirectory
+    {
+        /// <summary>
+        ///     Gets the dedicated directory where the given plugin is installed or should be installed.
+        /// </summary>
+        /// <param name="pluginIdentity">
+        ///     The identity of the plugin.
+        /// </param>
+        /// <returns>
+        ///     The directory where the given plugin is installed or should be installed.
```
mslukebo, line 22: I wonder if we want these methods to be returning _relative_ paths. That way the implementation doesn't need to prepend some "root directory," but rather the consumers do. Thoughts?

helenkzhang: Returning relative paths does make it more flexible but this interface has a lot of consumers (storage, loader, remover, checksum calculator and validator) and all of them will need to know about the plugins system root and append relative paths if we make this change. And these components are supposed to be sharing one root directory. Would it be better to just have `IPluginsStorageDirectory` to store the root?

mslukebo: That's a good point. I'd say leave it as-is and have this class store the root

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/Checksum/SHA256DirectoryChecksumCalculator.cs
helenkzhang: Updated

mslukebo (approved):

PR #289, opened 2023-04-11
Metadata Equality

helenkzhang:
- Implements `IEquatable`
- Override object `Equals` and `GetHashCode`
- Add `PluginIdentity` property in `PluginMetadata`

### DIFF
--- a/src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/EnumerableExtensions.cs
+++ b/src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/EnumerableExtensions.cs
@@ -1,0 +1,71 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Performance.Toolkit.Plugins.Core
+{
+    /// <summary>
+    ///     Provides extension methods for <see cref="IEnumerable{T}"/>.
+    /// </summary>
+    internal static class EnumerableExtensions
+    {
+        /// <summary>
+        ///     Determines whether two sequences are equal by comparing their elements by using the default equality comparer for their type.
+        ///     If both sequences are null, they are considered equal.
+        /// </summary>
+        /// <typeparam name="T">
+        ///     The type of the elements of the input sequences.s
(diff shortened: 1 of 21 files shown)


src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/EnumerableExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Performance.Toolkit.Plugins.Core
+{
+    /// <summary>
+    ///     Provides extension methods for <see cref="IEnumerable{T}"/>.
+    /// </summary>
+    internal static class EnumerableExtensions
```
mslukebo, line 13: Wonder if this should go into SDK?

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/Packaging/Metadata/ProcessingSourceMetadata.cs
```diff
@@ -67,5 +69,62 @@ public sealed class ProcessingSourceMetadata
...
+
+            if (this.AvailableTables != null)
+            {
+                foreach (TableMetadata table in this.AvailableTables)
+                {
+                    result = HashCodeUtils.CombineHashCodeValues(result, table?.GetHashCode() ?? 0);
+                }
+            }
+
+            if (this.SupportedDataSources != null)
```
mslukebo, line 119: Wonder if we should make a CombineHashCodeValues that takes in an `IEnumerable` that does this null check + loop

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/EnumerableExtensions.cs
helenkzhang: It should. But for now we want to put everything in PluginsSystem folder. We can do a refactor once plugins stuff goes official.

helenkzhang: Same for the other comment :)

mslukebo (approved):

PR #292, opened 2023-04-13
Clean ups

helenkzhang:
- Remove Cli csproj
- Update IPluginsDiscoverer to IPluginsDiscoveryOrchestrator to differentiate it from Core.IPluginDiscoverer
- Update `Equals` to use Ordinal for all metadata classes
- Update Plugins Glossary doc


jmaxson-ms (approved):
:shipit:

PR #302, opened 2023-05-31
User/jmaxson/filtered plugin loading

jmaxson-ms:
Provides a way for a plugin to specify that files/folder should be skipped when loading the plugin's folder. When the file 'PluginLoadExclusions" is found in a folder, each line of that file is interpreted as a file or folder to exclude from the discovery process.

Runtime code related to extension discovery that wasn't being used was removed.

Tests have been updated to test exclusion of a file and a folder.
A test has been added to test successfully loading from a folder.
Logging has been added when exclusions are enabled.

### DIFF
--- a/src/Microsoft.Performance.SDK.Runtime/Discovery/AssemblyExtensionDiscovery.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/Discovery/AssemblyExtensionDiscovery.cs
@@ -192,11 +192,16 @@ namespace Microsoft.Performance.SDK.Runtime.Discovery
                     HashSet<string> toIgnore = GetToIgnore(directoryPath, this.logger);
 
                     IEnumerable<string> subDirectories = this.FindFiles
-                        .EnumerateFolders(directoryPath)
-                        .Where(x => !toIgnore.Contains(Path.GetDirectoryName(x)));
+                        .EnumerateFolders(directoryPath);
 
                     foreach (string subDirectory in subDirectories)
                     {
+                        if (toIgnore.Contains(Path.GetDirectoryName(subDirectory)))
+                        {
+                            this.logger?.Verbose("Process assemblies: excluding directory '{0}'.", subDirectory);
+                            continue;
+                        }
+
                         if (!foldersSearched.Contains(subDirectory))
                         {
                             foldersToSearch.Push(subDirectory);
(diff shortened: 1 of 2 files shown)


src/Microsoft.Performance.SDK.Runtime/Discovery/AssemblyExtensionDiscovery.cs
```diff
@@ -23,6 +22,9 @@ namespace Microsoft.Performance.SDK.Runtime.Discovery
     public class AssemblyExtensionDiscovery
         : IExtensionTypeProvider
     {
+        public static readonly string ExclusionsFilename = "PluginLoadExclusions";
```
mslukebo, line 25: does this need to be public?

jmaxson-ms: don't think so. I can change it.

src/Microsoft.Performance.SDK.Runtime/Discovery/AssemblyExtensionDiscovery.cs
```diff
@@ -452,22 +409,65 @@ public bool ProcessAssemblies(IEnumerable<string> directoryPaths, out ErrorInfo
             return true;
         }
 
+        private static HashSet<string> GetExclusionValues(string targetDirectory, ILogger logger)
+        {
+            Debug.Assert(!string.IsNullOrWhiteSpace(targetDirectory), $"{nameof(targetDirectory)} is null.");
+            Debug.Assert(Directory.Exists(targetDirectory), $"{nameof(targetDirectory)} is invalid.");
+
+            string configurationFile = Directory
```
mslukebo, line 417: Does this really need to call `GetFiles`? I think a `Path.Combine` followed by a `File.Exists` would suffice

src/Microsoft.Performance.SDK.Runtime/Discovery/AssemblyExtensionDiscovery.cs
```diff
@@ -452,22 +409,65 @@ public bool ProcessAssemblies(IEnumerable<string> directoryPaths, out ErrorInfo
...
+                .GetFiles(targetDirectory, AssemblyExtensionDiscovery.ExclusionsFilename)
+                .FirstOrDefault();
+
+            if (configurationFile != null)
+            {
+                try
+                {
+                    IEnumerable<string> lines = File
```
mslukebo, line 425: Should we add a check to make sure lines do not contain directory separator characters? Maybe we should log a warning and ignore any entries that do

src/Microsoft.Performance.SDK.Runtime/Discovery/AssemblyExtensionDiscovery.cs
```diff
@@ -452,22 +409,65 @@ public bool ProcessAssemblies(IEnumerable<string> directoryPaths, out ErrorInfo
...
+        private static HashSet<string> GetExclusionValues(string targetDirectory, ILogger logger)
+        {
+            Debug.Assert(!string.IsNullOrWhiteSpace(targetDirectory), $"{nameof(targetDirectory)} is null.");
+            Debug.Assert(Directory.Exists(targetDirectory), $"{nameof(targetDirectory)} is invalid.");
+
+            string configurationFile = Directory
+                .GetFiles(targetDirectory, AssemblyExtensionDiscovery.ExclusionsFilename)
+                .FirstOrDefault();
+
+            if (configurationFile != null)
```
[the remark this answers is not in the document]
jmaxson-ms: I did it this way so that the exception would just fall through.

mslukebo (approved):

helenkzhang (approved):

PR #308, opened 2023-06-20
Plugin system version should bump if sdk changes

helenkzhang:


mslukebo:
What's the reason for this?

mslukebo (approved):

helenkzhang:
> > What's the reason for this?
> 
> When there's a change in the SDK, the version of PluginsSystem should be bumped. The pathFilters should be set to include its parent folder which is `src`.

Chatted offline with @mslukebo Because PluginsSystem has dependencies on the SDK, changes in the SDK should be treated as a new commit in the PluginsSystem which should result in a new version.

jmaxson-ms:
> > > What's the reason for this?
> > 
> > 
> > When there's a change in the SDK, the version of PluginsSystem should be bumped. The pathFilters should be set to include its parent folder which is `src`.
> 
> Chatted offline with @mslukebo Because PluginsSystem has dependencies on the SDK, changes in the SDK should be treated as a new commit in the PluginsSystem which should result in a new version.

I noticed we excluded the solution file. Should we exclude anything else, like test folder changes?

jmaxson-ms (approved):

helenkzhang:
> 



> > > > What's the reason for this?
> > > 
> > > 
> > > When there's a change in the SDK, the version of PluginsSystem should be bumped. The pathFilters should be set to include its parent folder which is `src`.
> > 
> > 
> > Chatted offline with @mslukebo Because PluginsSystem has dependencies on the SDK, changes in the SDK should be treated as a new commit in the PluginsSystem which should result in a new version.
> 
> I noticed we excluded the solution file. Should we exclude anything else, like test folder changes?


> > > > What's the reason for this?
> > > 
> > > 
> > > When there's a change in the SDK, the version of PluginsSystem should be bumped. The pathFilters should be set to include its parent folder which is `src`.
> > 
> > 
> > Chatted offline with @mslukebo Because PluginsSystem has dependencies on the SDK, changes in the SDK should be treated as a new commit in the PluginsSystem which should result in a new version.
> 
> I noticed we excluded the solution file. Should we exclude anything else, like test folder changes?
> 
> edit: looking through the folders, it seems like it might be easier to create a list of folders we want to include. do we update when the engine changes? what about the runtime?

Chatted offline. We should fine tune the versioning  

> > > > What's the reason for this?
> > > 
> > > 
> > > When there's a change in the SDK, the version of PluginsSystem should be bumped. The pathFilters should be set to include its parent folder which is `src`.
> > 
> > 
> > Chatted offline with @mslukebo Because PluginsSystem has dependencies on the SDK, changes in the SDK should be treated as a new commit in the PluginsSystem which should result in a new version.
> 
> I noticed we excluded the solution file. Should we exclude anything else, like test folder changes?
> 
> edit: looking through the folders, it seems like it might be easier to create a list of folders we want to include. do we update when the engine changes? what about the runtime?

Chatted offline. Created an issue to address in another PR https://github.com/microsoft/microsoft-performance-toolkit-sdk/issues/309

PR #311, opened 2023-07-05
PluginsSystem: Add ability to validate plugins before they are used

mslukebo:


src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Validation/PluginsSystemOperation.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// </summary>
+    public enum PluginsSystemOperation
+    {
+        /// <summary>
+        ///     One or more plugins are attempting to be installed.
+        /// </summary>
+        Install,
+
+        /// <summary>
+        ///  One or more plugins are attempting to be loaded..
```
helenkzhang, line 17: nit: Remove the extra period.

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/FileBackedPluginsInstaller.cs
```diff
@@ -95,7 +77,9 @@ public sealed class FileBackedPluginsInstaller
             this.installedPluginValidator = installedPluginValidator;
             this.installedPluginStorage = installedPluginStorage;
             this.pluginPackageReader = packageReader;
+            this.validator = validator;
```
helenkzhang, line 80: Rename this one too?

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/PluginsSystemCreationOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+        ///     Gets or sets additional <see cref="IPluginValidator"/>s that will be used to validate
+        ///     plugins during plugin discovery, installation, and loading.
+        /// </summary>
+        public IReadOnlyCollection<IPluginValidator> AdditionalValidators { get; set; } =
+            Array.Empty<IPluginValidator>();
+
+        /// <summary>
+        ///     Gets or sets an <see cref="IInvalidPluginsGate"/> to use when determining whether
+        ///     to continue a plugins system operation despite plugin validation errors.
+        ///     If a value is not set, operations will *never* continue if validation errors occur.
```
jmaxson-ms, line 32: > If a value is not set, operations will *never* continue if validation errors occur. [](http://example.com/codeflow?start=15&length=84)

Maybe something like this instead?

The default behavior is that operations will *never* continue if validation errors occur.

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Validation/IPluginValidator.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <param name="pluginMetadata">
+        ///     The metadata for the plugin attempting to be used.
+        /// </param>
+        /// <param name="errorInfos">
+        ///     <see cref="ErrorInfo"/>s describing why <paramref name="pluginMetadata"/> is invalid.
+        /// </param>
+        /// <returns>
+        ///     <c>true</c> if <paramref name="pluginMetadata"/> is valid, <c>false</c> otherwise.
+        /// </returns>
+        bool IsValid(PluginMetadata pluginMetadata, out ErrorInfo[] errorInfos);
```
jmaxson-ms, line 26: >bool IsValid(PluginMetadata pluginMetadata, out ErrorInfo[] errorInfos); [](http://example.com/codeflow?start=8&length=72)

Do we ever use the 'bool' result? Should we just return ErrorInfo[] instead? (And rename the method if we do this.)

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Validation/PluginsSystemOperation.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.Performance.Toolkit.Plugins.Runtime.Validation
+{
+    /// <summary>
+    ///     Represents an operation within the plugins system that requires plugin validation.
+    /// </summary>
+    public enum PluginsSystemOperation
+    {
+        /// <summary>
+        ///     One or more plugins are attempting to be installed.
```
jmaxson-ms, line 12: >One or more plugins are attempting to be installed. [](http://example.com/codeflow?start=16&length=51)

nit: sounds like the plugin is doing the work.

Another comment in another file has this:

"The <see cref="IPluginValidator"/> for validating a plugin that is being installed."

I think we might want to change this, and the description below, to be more consistent. Maybe something like:

Attempting to install one or more plugins.

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Validation/IPluginValidator.cs
jmaxson-ms: I read the PluginsDiscoveryOrchestrator first and it wasn't using this return value, but FileBackedPluginsInstaller is. Still seems kinda weird. I still prefer to always just return ErrorInfo[] and then check for null/!any().

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Loading/FileSystemInstalledPluginLoader.cs
```diff
@@ -65,10 +79,21 @@ public async Task<bool> LoadPluginAsync(InstalledPlugin installedPlugin)
 
             if (!await this.installedPluginValidator.ValidateInstalledPluginAsync(installedPlugin.Info))
             {
-                this.logger.Error($"Plugin {pluginIdentity} is not valid. It will not be loaded.");
+                this.logger.Error($"Plugin {pluginIdentity} is not valid installed plugin. It will not be loaded. Has it been corrupted?");
```
jmaxson-ms, line 82: >Has it been corrupted? [](http://example.com/codeflow?start=114&length=22)

Should we add some hash information so that we can determine whether it's been corrupted? Seems like a strange question to ask in a log comment.

mslukebo: I updated the implementation to log that message

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Validation/IPluginValidator.cs
```diff
@@ -0,0 +1,26 @@
...
+        ///     Validates a plugin based on its metadata, returning <see cref="ErrorInfo"/>s for
+        ///     any validation issues found.
+        /// </summary>
+        /// <param name="pluginMetadata">
+        ///     The metadata for the plugin to validate.
+        /// </param>
+        /// <returns>
+        ///     <see cref="ErrorInfo"/>s describing why <paramref name="pluginMetadata"/> is invalid.
+        /// </returns>
+        ErrorInfo[] ValidationErrors(PluginMetadata pluginMetadata);
```
jmaxson-ms, line 24: >ValidationErrors [](http://example.com/codeflow?start=20&length=16)

sorry: seems like this should be named with a verb. Validate() or ValidatePlugin()

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Validation/IPluginValidator.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        /// <summary>
+        ///     Validates a plugin based on its metadata, returning <see cref="ErrorInfo"/>s for
+        ///     any validation issues found.
+        /// </summary>
+        /// <param name="pluginMetadata">
+        ///     The metadata for the plugin to validate.
+        /// </param>
+        /// <returns>
+        ///     <see cref="ErrorInfo"/>s describing why <paramref name="pluginMetadata"/> is invalid.
```
jmaxson-ms, line 22: ><see cref="ErrorInfo"/>s describing why <paramref name="pluginMetadata"/> is invalid. [](http://example.com/codeflow?start=16&length=85)

or null for a valid plugin?

jmaxson-ms: oh, looks like an empty array.

helenkzhang (approved):

jmaxson-ms (approved):
:shipit:

PR #312, opened 2023-07-06
Add metadata versioning

helenkzhang:
- Added `SchemaVersion` to DTO types and set to 0.1
- Added `PublishedOn` to `AvailablePluginInfo`
- Added `ProjectUrl` to `PluginMetadata`

This is the temporary solution to prevent future changes to the schemas breaking backward compatibility, down the road we need to work on https://github.com/microsoft/microsoft-performance-toolkit-sdk/issues/304


src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/AvailablePluginInfo.cs
```diff
@@ -18,13 +18,32 @@ public sealed class AvailablePluginInfo
...
+        /// </param>
         [JsonConstructor]
         public AvailablePluginInfo(
             PluginMetadata metadata,
             PluginSource source,
             ulong packageSize,
             Uri packageUri,
-            Guid fetcherResourceId)
+            Guid fetcherResourceId,
+            DateTimeOffset publishedOn)
```
anramanath, line 46: Is DateTimeOffset better than DateTime here?

anramanath: https://stackoverflow.com/questions/4331189/datetime-vs-datetimeoffset

Great read, looks good with DateTimeOffset.

jmaxson-ms (approved):
:shipit:

anramanath (approved):

PR #313, opened 2023-07-25
Plugins: allow resetting registry if it becomes corrupted

mslukebo:
When deserializing the plugins registry, the caller should have a way to recover from a corrupt file. This PR adds a `ResetInstalledPlugins` method to the `IPluginsInstaller` which will effectively delete the registry file.

Additionally, this PR
1. Documents that `IPluginsInstaller` can throw `RepositoryCorruptedException` and `RepositoryDataAccessException`
2. Updates the repository to throw `RepositoryCorruptedException` when a `JsonException` is encountered (previously, this threw `RepositoryDataAccessException`, which did not make sense to me)
3. Also throws `RepositoryCorruptedException` when an `ArgumentNullException` is encountered during deserialization

The last change is necessary because System.Text.Json will not throw a JsonException if a deserialized class is missing members (although, this functionality is coming in .NET 8). Instead, the deserialized classes are using `Guard.NotNull` clauses that throw `ArgumentNullException` when a deserialized constructor parameter is null (i.e. not present in the JSON).

### DIFF
--- a/src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Registry/FileBackedPluginRegistry.cs
+++ b/src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Registry/FileBackedPluginRegistry.cs
@@ -238,7 +238,17 @@ namespace Microsoft.Performance.Toolkit.Plugins.Runtime
         /// <inheritdoc/>
         public Task Reset(CancellationToken cancellationToken)
         {
-            return Task.Run(() => File.Delete(this.registryFilePath), cancellationToken);
+            return Task.Run(() =>
+            {
+                try
+                {
+                    File.Delete(this.registryFilePath);
+                }
+                catch (IOException e)
+                {
+                    throw new RepositoryDataAccessException("Failed to delete plugin registry.", e);
+                }
+            }, cancellationToken);
         }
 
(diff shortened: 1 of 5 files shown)


helenkzhang (dismissed):

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Registry/FileBackedPluginRegistry.cs
```diff
@@ -235,6 +235,12 @@ public async Task<IReadOnlyCollection<InstalledPluginInfo>> GetAllAsync(Cancella
             return await this.fileDistributedLock.AcquireAsync(timeout, cancellationToken);
         }
 
+        /// <inheritdoc/>
+        public Task Reset(CancellationToken cancellationToken)
+        {
+            return Task.Run(() => File.Delete(this.registryFilePath), cancellationToken);
```
jmaxson-ms, line 241: >return Task.Run(() => File.Delete(this.registryFilePath), cancellationToken); [](http://example.com/codeflow?start=12&length=77)

do we want to catch IO exceptions and throw a something like RepositoryDataAccessException instead?

jmaxson-ms (approved):
:shipit:

helenkzhang (approved):

PR #317, opened 2023-08-23
Add Plugin CLI Tool

helenkzhang:
https://github.com/microsoft/microsoft-performance-toolkit-sdk/blob/1f49fb2d0d4ba80d0cea1fc9a4b4041f93643481/documentation/Plugins-System-Preview-Only/PluginTool-CLI.md


src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Program.cs
```diff
@@ -0,0 +1,226 @@
...
+            try
+            {
+                return result.MapResult(
+                    (PackOptions opts) => RunPack(opts, serviceProvider),
+                    (MetadataGenOptions opts) => RunMetadataGen(opts, serviceProvider),
+                    errs => HandleParseError(errs, serviceProvider));
+            }
+            catch (ArgumentValidationException ex)
+            {
+                logger.LogError("Invalid arguments. {0}", ex.Message);
```
mslukebo, line 66: To me, having all these exceptions is a code smell. Are these truly "exceptional?" For instance, invalid arguments aren't an exception... we should expect and handle that case gracefully, returning early with a useful error message/code. Same with other invalid inputs (manifest, plugin content, etc.).

The issue I have with this pattern is the program's entry point is essentially an uber-class, needing to know about everything that could possibly go wrong while the program is executing. As more things are checked, this list grows.

I personally would get rid of these exceptions and just use facades to abstract out individual steps

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Program.cs
```diff
@@ -0,0 +1,226 @@
...
+            }
+
+            logger.LogInformation($"Successfully created plugin contents metadata at {validDestContentsMetadataFileName}.");
+            
+            return 0;
+        }
+
+        private static void PackageGenCommon(PackageGenCommonOptions options, ServiceProvider serviceProvider, out ProcessedPluginDirectory processedDir, out PluginMetadata metadata)
+        {
+            ISourceFilesProcessor processor = serviceProvider.GetRequiredService<ISourceFilesProcessor>();
```
mslukebo, line 191: This is called the "Service Locator" pattern, which in modern times is considered a code smell 😄. This pattern undermines the point of using dependency injection.

Instead, you should be creating classes for each of the different actions that can be performed. Each class should have a set of dependencies in its constructor that get automatically satisifed via dependency injection.

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Constants.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Performance.Toolkit.Plugins.Cli
+{
+    public static class Constants
```
mslukebo, line 6: This file needs to be cleaned up

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Microsoft.Performance.Toolkit.Plugins.Cli.csproj
```diff
@@ -0,0 +1,35 @@
...
+	<ToolCommandName>plugin</ToolCommandName>
+	<PackageOutputPath>./nupkg</PackageOutputPath>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="CommandLineParser" Version="2.9.1" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="7.0.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
```
mslukebo, line 19: Do we need Newtonsoft?

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Options/PackageGenCommonOptions.cs
```diff
@@ -0,0 +1,73 @@
...
+            "overwrite",
+            Required = false,
+            HelpText = "Indicates that the destination file should be overwritten if it already exists.")]
+        public bool Overwrite { get; } = false;
+
+        [Option(
+            'm',
+            "manifest",
+            Required = false,
+            HelpText = "Path to the plugin manifest file. If not specified, the program will attempt to find the manifest in the source directory.")]
```
mslukebo, line 39: Thoughts on just requiring this?

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Options/PackageGenCommonOptions.cs
```diff
@@ -0,0 +1,73 @@
...
+        public string? ManifestFilePath { get; }
+
+        internal string SourceDirectoryFullPath { get; private set; }
+
+        internal string? ManifestFileFullPath { get; private set; }
```
mslukebo, line 44: IMO, it's better to separate out the "parse result" with "actual representation the app wants to work with" into two separate classes. The idea being you have a single class whose only job is to contain the parsed result, Then you have a way to transform that class into whatever representation you actually want that have these above properties

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Metadata/MetadataGenerator.cs
```diff
@@ -0,0 +1,42 @@
...
+        private readonly ILogger<MetadataGenerator> logger;
+
+        public MetadataGenerator(ILogger<MetadataGenerator> logger)
+        {
+            this.logger = logger;
+        }
+
+        public PluginMetadata Generate(ProcessedPluginDirectory pluginDirectory, PluginManifest manifest)
+        {
+            PluginIdentity identity = new(manifest.Identity.Id, manifest.Identity.Version);
```
mslukebo, line 25: We should probably augment a lot of these methods with logger.Verbose calls so we can track what happens

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Validation/ProcessedPluginDirectory.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Performance.Toolkit.Plugins.Core.Metadata;
+
+namespace Microsoft.Performance.Toolkit.Plugins.Cli.Validation
+{
+    public record ProcessedPluginDirectory(
+        string FullPath,
+        string? ManifestFilePath,
```
mslukebo, line 10: Why is this nullable? I feel like this shouldn't be part of this record at all

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/IPack.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Performance.Toolkit.Plugins.Cli.Options;
+
+namespace Microsoft.Performance.Toolkit.Plugins.Cli
+{
+    internal interface IPack
```
mslukebo, line 8: Instead of having a bunch of interfaces, you can just create a single generic one:

```cs
internal interface IPluginToolkitAction<T>
    where T : PackageGenCommonOptions
{
    int Run(T options);
}
```

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/IPluginContentsProcessor.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Performance.Toolkit.Plugins.Cli.Options;
+
+namespace Microsoft.Performance.Toolkit.Plugins.Cli
+{
+    internal interface IPluginContentsProcessor
+    {
+        public ProcessedPluginContents Process(PackageGenCommonOptions options);
```
mslukebo, line 10: if all options extend `PackageGenCommonOptions` would it make sense to have a base handler class instead of a new interface here?

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Console/PluginsCli.cs
```diff
@@ -0,0 +1,185 @@
...
+                {
+                    this.logger.LogError($"The directory '{outputDir}' does not exist. Please provide a valid directory path or create the directory and try again.");
+                }
+            }
+
+            result = new PackArgs(resultBase, outputFileFullPath);
+            return true;
+        }
+
+        private bool TryGetValidatdMetadataGenArgs(MetadataGenOptions options, out MetadataGenArgs? result)
```
mslukebo, line 119: I think the `TryGetValidatedXArgs` should all happen in different files. They could happen inside the ICommands themselves. I don't think this class needs to have any specific knowledge of specific options, though

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Utils.cs
```diff
@@ -0,0 +1,23 @@
+﻿
+namespace Microsoft.Performance.Toolkit.Plugins.Cli
+{
+    public static class Utils
+    {
+        public static string GetValidDestFileName(string file)
```
mslukebo, line 6: Nit: the name of this method doesn't explain what it does. What does "Valid" mean in this context?

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Processing/PluginArtifacts.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// <summary>
+    ///     Represents the artifacts of a plugin.
+    /// </summary>
+    /// <param name="SourceDirectoryFullPath">
+    ///     The full path to the directory containing the plugin binaries.
+    /// </param>
+    /// <param name="ManifestFileFullPath">
+    ///     The full path to the manifest file, if it's not in the source directory. Otherwise, <c>null</c>.
+    /// </param>
+    internal record PluginArtifacts(
```
mslukebo, line 15: Instead of using null, I would say just have a boolean value `bool SourceDirectoryContainsManifestFile`

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Console/PluginsCli.cs
helenkzhang: I think they should happen outside of ICommands because ICommands are not supposed to understand XOptions (cli framework dependent types) directly. These methods are transforming the cli specific options to "app" args while also validating these cli inputs. It makes sense to validate them in ICommands but not transform them. Any suggestions?

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Processing/PluginArtifacts.cs
mslukebo: Make it not nullable and just always have the path (regardless of whether or not it's in the source directory)

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Console/PluginsCli.cs
mslukebo: You can have an `IOptionsValidator<TOptions, TArgs>` that the commands depend on?

mslukebo: This way you can also unit test the validation logic in a single class

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Processing/PluginArtifacts.cs
helenkzhang: Well if it's in the source directory, we won't know the path to the manifest file until we call `TryProcess()`. The assumption for now is that manifest can be placed anywhere in the source directory. TryProcess would search for the name of the manifest in the source directory. To make things easier, we can enforce that pluginManifest.json must be at the root level.

mslukebo: Another approach is to instead make an interface `IManifestLocator` that has a single methof `string Locate()`. If it's within the source directory, we pass in a `SourceDirectoryManifestLocator`. Otherwise we pass in an implenetation that spits out the location provided on the command line

mslukebo: Basically, delegate the work of knowing how to extract the manifest location to another class (that we can unit test in isolation)

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Commands/Pack/PackOptionsValidator.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        public PackOptionsValidator(ILogger<PackOptionsValidator> logger)
+            : base(logger)
+        {
+        }
+
+        public override bool TryValidate(PackOptions cliOptions, out PackArgs? validatedAppArgs)
+        {
+            validatedAppArgs = null;
+            if (!TryValidateCommonOptions(cliOptions, out PackGenCommonArgs validatedCommonArgs))
```
mslukebo, line 22: Instead of having TryValidate be abstract and having each implementation call TryValidateCommonOptions, it might be better to create two methods on `PackGenCommonOptionsValidator`

```cs
public bool TryValidate(TOptions options, out TArgs args)
{
    if (TryValidateCommonOptions(options, out PackGenCommonArgs validatedCommonArgs))
    {
        return TryValidateCore(options, validatedCommonArgs, out args);
    }
    else
    {
        args = null;
        return false;
    }
}

protected abstract bool TryValidateCore(TOptions options, PackGenCommonArgs validatedAppArgs , out TArgs args);
```

This follows the pattern we do in the rest of the SDK

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Manifest/SourceDirectoryManifestLocator.cs
```diff
@@ -0,0 +1,48 @@
...
+            ILogger<SourceDirectoryManifestLocator> logger)
+        {
+            this.sourceDir = sourceDirectory;
+            this.logger = logger;
+        }
+
+        public bool TryLocate(out string? manifestFilePath)
+        {
+            manifestFilePath = null;
+            foreach (string file in Directory.EnumerateFiles(this.sourceDir, "*", SearchOption.AllDirectories))
```
mslukebo, line 25: Instead of using * and using a for loop, can't you just pass in `Constants.BundledManifestName` as the filter and do something like

```cs
var matchedFiles = Directory.EnumerateFiles(this.sourceDir, Constants.BundledManifestName, SearchOption.AllDirectories).ToList();

if (matchedFiles.Count == 0)
{
    this.logger.LogError($"Directory does not contain {Constants.BundledManifestName} as expected: {this.sourceDir}.");
    return false;
}
else if (matchedFiles.Count > 1)
{
    this.logger.LogError($"Directory contains multiple manifests: {manifestFilePath}, {file}. Only one manifest is allowed.");
    return false;
}
else
{
    manifestFilePath = matchedFiles.First();
    return true;
}

File: src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Commands/Pack/PackCommand.cs [hunk 16]
```diff
@@ -0,0 +1,80 @@
...
+        {
+            this.manifestLocatorFactory = manifestLocatorFactory;
+            this.sourceProcessor = sourceProcessor;
+            this.packageBuilder = packageBuilder;
+            this.logger = logger;
+        }
+
+        public int Run(PackArgs args)
+        {
+            IManifestLocator manifestLocator = this.manifestLocatorFactory.Create(args);
```
mslukebo, line 34: I'm noticing lines 34 through 46 are copy/pasted. Should we extract out a common dependency?

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Processing/PluginArtifacts.cs
helenkzhang: Yes

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Manifest/SourceDirectoryManifestLocator.cs
helenkzhang: Updated

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Commands/Pack/PackCommand.cs
helenkzhang: Introduced a base command instead.

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Commands/Common/PackGenCommonCommand.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                return 1;
+            }
+
+            return RunCore(args, processedSource!);
+        }
+
+        protected abstract int RunCore(TArgs args, ProcessedPluginResult processedSource);
+
+        protected bool TryGetProcessedPluginResult(PackGenCommonArgs args, out ProcessedPluginResult? processedPluginResult)
```
mslukebo, line 37: nit: this can be made private now

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Commands/Common/PackGenCommonCommand.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        public int Run(TArgs args)
+        {
+            if (!TryGetProcessedPluginResult(args, out ProcessedPluginResult? processedSource))
+            {
+                return 1;
+            }
+
+            return RunCore(args, processedSource!);
```
mslukebo, line 32: another nit:

Instead of using `!` everywhere, we can decorate `out` parameters with https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullwhenattribute?view=net-7.0

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Commands/Common/PackGenCommonOptionsValidator.cs
```diff
@@ -0,0 +1,70 @@
...
+                validatedAppArgs = null;
+                return false;
+            }
+
+            return TryValidateCore(cliOptions, validatedCommonArgs, out validatedAppArgs);
+        }
+
+        protected abstract bool TryValidateCore(TOptions options, PackGenCommonArgs validatedCommonAppArgs, out TArgs? validatedAppArgs);
+
+        protected bool TryValidateCommonOptions(
```
mslukebo, line 34: can be made private?

mslukebo (dismissed):

mslukebo (dismissed):

trgibeau (dismissed):

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Processing/PluginArtifactsProcessor.cs
```diff
@@ -109,10 +109,11 @@ public bool TryProcess(PluginArtifacts artifacts, [NotNullWhen(true)] out Proces
                     {
                         assembly = Assembly.LoadFrom(file);
```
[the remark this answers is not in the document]
mslukebo: edit: I see we use it below... why do we call this here?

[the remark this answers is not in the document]
helenkzhang: The PluginsLoader doesn't expose the SDK version of the plugin in the loading result. It does validate the version using VersionChecker but doesn't save the version. We probably should add a TODO to make pluginsloader save that. It's not trivial.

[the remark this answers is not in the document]
mslukebo: Hmm this is something we're going to have to do sooner rather than later. We can't just bail on these files, because e.g. the tool can be ran on an x64 machine for an arm64 plugin (although, plugins should _really_ be compiled for AnyCPU...).

We use [MetadataLoadContext](https://learn.microsoft.com/en-us/dotnet/standard/assembly/inspect-contents-using-metadataloadcontext) in the runtime to inspect the referenced SDK without loading the assembly. It's built for this type of scenario. Maybe we need to use that here? Or figure out a way to use the runtime to get this info?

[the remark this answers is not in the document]
helenkzhang: Tested `MetadataLoadContext.Load` - it will still throw if when loading native dlls.

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Commands/Pack/PackCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            string tmpFile = Path.GetTempFileName();
+            try
+            {
+                this.packageBuilder.Build(processedSource!, tmpFile);
+                File.Move(tmpFile, destFilePath, args.Overwrite);
+            }
+            catch (Exception ex)
+            {
+                this.logger.LogError($"Failed to create plugin package at {destFilePath} due to an exception {ex.Message}");
```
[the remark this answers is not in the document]
helenkzhang: Good catch. We don't want to log the whole exception. I updated the other file.

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Commands/MetadataGen/MetadataGenCommand.cs
```diff
@@ -0,0 +1,86 @@
...
+        protected override int RunCore(MetadataGenArgs args, ProcessedPluginResult processedSource)
+        {
+            PluginMetadata metadata = processedSource.Metadata;
+            PluginContentsMetadata contentsMetadata = processedSource.ContentsMetadata;
+
+            bool outputSpecified = args.OutputDirectoryFullPath != null;
+            string? outputDirectory = outputSpecified ? args.OutputDirectoryFullPath : Environment.CurrentDirectory;
+
+            string destMetadataFileName = Path.Combine(outputDirectory!, $"{metadata.Identity}-{PackageConstants.PluginMetadataFileName}");
+            string validDestMetadataFileName = outputSpecified && args.Overwrite ?
```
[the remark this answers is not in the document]
helenkzhang: Added a bool. If we overwrite, we don't need to get alternative paths basically.

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Manifest/ManifestLocatorFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        /// <inheritdoc />
+        public IManifestLocator Create(PackGenCommonArgs args)
+        {
+            if (args.ManifestFileFullPath != null)
+            {
+                return new CommandLineManifestFileLocator(args.ManifestFileFullPath);
+            }
+            else
```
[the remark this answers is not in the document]
helenkzhang: Updated

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Processing/PluginArtifactsProcessor.cs
```diff
@@ -0,0 +1,287 @@
...
+            }
+
+            string manifestFilePath = artifacts.ManifestFileFullPath;
+
+            try
+            {
+                if (!this.manifestValidator.IsValid(manifestFilePath, out List<string> validationMessages))
+                {
+                    string errors = string.Join(Environment.NewLine, validationMessages);
+                    this.logger.LogWarning($"Manifest file failed some json schema format validation checks: \n{errors}");
```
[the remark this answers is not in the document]
helenkzhang: The next warning will automatically be in a new line.

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Commands/Common/PackGenCommonCommand.cs
```diff
@@ -0,0 +1,68 @@
...
+            this.artifactsProcessor = artifactsProcessor;
+            this.logger = logger;
+        }
+
+        /// <inheritdoc />
+        public int Run(TArgs args)
+        {
+            if (!TryGetProcessedPluginResult(args, out ProcessedPluginResult? processedSource))
+            {
+                return 1;
```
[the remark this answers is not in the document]
helenkzhang: Will keep as is for now since we only have 0 and 1 as exit codes and it's not really used anywhere. We can add a constant file in the future when we have more codes that represent specific errors.

src/Microsoft.Performance.SDK.Runtime/TrackingVersionChecker.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// </param>
+        /// <returns>
+        ///     <c>true</c> if the version is supported; <c>false</c> otherwise.
+        /// </returns>
+        public override bool IsVersionSupported(SemanticVersion candidateVersion)
+        {
+            bool supported = base.IsVersionSupported(candidateVersion);
+            if (supported)
+            {
+                this.verifiedVersions.Add(candidateVersion);
```
mslukebo, line 51: I think we want to always add this. If the tool is compiled against SDK 1.0, a plugin sing 1.1 should still be reported

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Commands/MetadataGen/MetadataGenOptionsValidator.cs
```diff
@@ -0,0 +1,53 @@
...
+                return false;
+            }
+
+            string? outputDirectoryFullPath = null;
+            if (cliOptions.OutputDirectory != null)
+            {
+                if (!Directory.Exists(cliOptions.OutputDirectory))
+                {
+                    this.logger.LogError($"Output directory '{cliOptions.OutputDirectory}' does not exist." +
+                        $"Please create it or omit the --output option to use the current directory.");
```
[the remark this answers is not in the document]
helenkzhang: Yeah we might want to support that but for this version will keep that it requires the client to create a valid directory. We need to properly handle the deletion of the created directory if something bad happens.

src/Microsoft.Performance.SDK.Runtime/TrackingVersionChecker.cs
helenkzhang: Updated

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Processing/PluginArtifactsProcessor.cs
```diff
@@ -77,27 +77,52 @@ public bool TryProcess(PluginArtifacts artifacts, [NotNullWhen(true)] out Proces
...
+                if (pluginSDKversion.Minor > cliSDKVersion.Minor)
+                {
+                    this.logger.LogError($"Plugin is built against SDK version {pluginSDKversion} but the sdk used in the CLI is {cliSDKVersion}. " +
+                        "The minor version of the SDK used in the CLI must be greater than or equal to the minor version of the SDK used to build the plugin. " +
+                        $"If your plugin does NOT use any features from SDK version {pluginSDKversion}, consider downgrading the plugin to use version {cliSDKVersion}. " +
+                        $"If your plugin does use features from SDK version {pluginSDKversion}, please update your CLI to a version that targets SDK version {pluginSDKversion} or later.");
+                }
+            }
+
+            if (!loadSuccess || !isVersionSupported)
```
mslukebo, line 120: super nit: we already return early if `!loadSuccess` above. To make it more clear that `!isVersionSupported` should return false, can we just add a `return false` after line 117?

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Processing/PluginArtifactsProcessor.cs
```diff
@@ -77,27 +77,52 @@ public bool TryProcess(PluginArtifacts artifacts, [NotNullWhen(true)] out Proces
...
+                    this.logger.LogError($"Plugin is built against SDK version {pluginSDKversion} but the sdk used in the CLI is {cliSDKVersion}. " +
+                        "The major version of the SDK used in the CLI must match the major version of the SDK used to build the plugin. " +
+                        "Please use the CLI that targets the same major version of the SDK as the plugin.");
+                }
+
+                if (pluginSDKversion.Minor > cliSDKVersion.Minor)
+                {
+                    this.logger.LogError($"Plugin is built against SDK version {pluginSDKversion} but the sdk used in the CLI is {cliSDKVersion}. " +
+                        "The minor version of the SDK used in the CLI must be greater than or equal to the minor version of the SDK used to build the plugin. " +
+                        $"If your plugin does NOT use any features from SDK version {pluginSDKversion}, consider downgrading the plugin to use version {cliSDKVersion}. " +
```
mslukebo, line 115: > consider downgrading the plugin to use version {cliSDKVersion}

I think more accurately, "consider downgrading the plugin to use version {cliSDKVersion} or lower."

Also we should call out that using the lowest minor version required will maximize compatibility of the plugin

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Processing/PluginArtifactsProcessor.cs
helenkzhang: Actually, we do not return early if `!loadSuccess` . See line 84. We just log the errorinfo and continue because if the failure is caused by incompatible sdk version, we want to log the messages properly.

mslukebo (approved):

anramanath (approved):

PR #328, opened 2023-09-26
User/jmaxson/loading fixes

jmaxson-ms:
When iterating over assemblies and looking for extension types, do not attempt to recreate types that have already been loaded. This can happen if an assembly exists in multiple directories.

The runtime was creating an ILogger which wrote to the console. This was a bug as, in some cases, the caller couldn't override this behavior. The default behavior (runtime) now is to not log if a logger is not specified (Logger.Null). Some methods overrides have been added to allow the caller to specify a logger to use.

### DIFF
--- a/src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Tables/TableExtensionReference.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Tables/TableExtensionReference.cs
@@ -134,6 +134,9 @@ namespace Microsoft.Performance.SDK.Runtime.Extensibility.DataExtensions.Tables
         /// <param name="candidateType">
         ///     Candidate <see cref="Type"/> for the <see cref="ITableExtensionReference"/>
         /// </param>
+        /// <param name="logger">
+        ///     Logs messages during reference creation.
+        /// </param>
         /// <param name="reference">
         ///     Out <see cref="ITableExtensionReference"/>
         /// </param>
@@ -146,15 +149,18 @@ namespace Microsoft.Performance.SDK.Runtime.Extensibility.DataExtensions.Tables
         /// </exception>
         internal static bool TryCreateReference(
             Type candidateType,
+            ILogger logger,
             out ITableExtensionReference reference)
         {
-            Guard.NotNull(candidateType, nameof(candidateType));
+            Debug.Assert(candidateType != null, $"{nameof(candidateType)} cannot be null.");
+            Debug.Assert(logger != null, $"{nameof(logger)} cannot be null.");
 
             reference = null;
 
             if (TableDescriptorFactory.TryCreate(
                     candidateType,
                     tableConfigSerializer,
+                    logger,
                     out var tableDescriptor,
                     out var tableBuildAction,
                     out var tableIsDataAvailableFunc))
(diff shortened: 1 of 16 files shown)


src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Tables/TableExtensionReference.cs
```diff
@@ -146,15 +149,18 @@ public TableDescriptor TableDescriptor
         /// </exception>
         internal static bool TryCreateReference(
             Type candidateType,
+            ILogger logger,
             out ITableExtensionReference reference)
         {
-            Guard.NotNull(candidateType, nameof(candidateType));
+            Debug.Assert(candidateType != null, $"{nameof(candidateType)} cannot be null.");
```
mslukebo, line 155: Why change these to debug asserts?

src/Microsoft.Performance.Toolkit.Engine/PluginSet.cs
```diff
@@ -355,7 +355,7 @@ public static PluginSet Load(string extensionDirectory)
                 var factory = new DataExtensionFactory();
                 repo = factory.CreateDataExtensionRepository();
 
-                var reflector = new DataExtensionReflector(assemblyDiscovery, repo);
+                var reflector = new DataExtensionReflector(assemblyDiscovery, repo, logger ?? Logger.Null);
```
mslukebo, line 358: On line 344, we call `new AssemblyLoader()`, which does `Runtime.Logger.Create<AssemblyLoader>`. We might wat to do the same thing there

src/Microsoft.Performance.SDK.Runtime/Extensibility/DataExtensions/Tables/TableExtensionReference.cs
jmaxson-ms: It's on an internal method. We only throw on publicly facing APIs. We Debug.Assert if it's internal to our code and only we should be calling it.

src/Microsoft.Performance.SDK.Runtime/SourceDataCookerReferenceFactory.cs
```diff
@@ -31,7 +32,37 @@ public static class SourceDataCookerReferenceFactory
...
+            ILogger logger,
+            out ISourceDataCookerReference reference)
+        {
+            Guard.NotNull(candidateType, nameof(candidateType));
+            Guard.NotNull(logger, nameof(logger));
+
+            return SourceDataCookerReference.TryCreateReference(candidateType, logger, out reference);
         }
     }
 }
```
helenkzhang, line 68: nit: Add new line at EOF

src/Microsoft.Performance.Toolkit.Engine/PluginSet.cs
jmaxson-ms: I'm actually going to do the reverse and use Logger.Null if not specified so that there's a consistent, don't use console by default, behavior.

mslukebo: That's what I meant 😄

helenkzhang (dismissed):

mslukebo (dismissed):

helenkzhang (dismissed):

PR #346, opened 2024-02-28
Engine.Create(IDataSource) disposes of DataSourceSet.

jmaxson-ms:
What's "funny" about this bug is that the comment in Engine.Dispose() directly references this case.

```
//
// we diposes the original set only if we own the data sources.
// this only occurs when the user uses one of the convenience
// methods that creates a DataSource internally, **e.g. Create(IDataSource).**
//

this.internalDataSourceSet.SafeDispose();
```

I added tests for each of the helper Create methods that should be disposing of the DataSourceSet.

### DIFF
--- a/src/Microsoft.Performance.Toolkit.Engine.Tests/TestCookers/Source123/Source123DataSource.cs
+++ b/src/Microsoft.Performance.Toolkit.Engine.Tests/TestCookers/Source123/Source123DataSource.cs
@@ -15,7 +15,8 @@ namespace Microsoft.Performance.Toolkit.Engine.Tests.TestCookers.Source123
        "Source123 for Runtime Tests")]
     [FileDataSource(Extension)]
     public sealed class Source123DataSource
-        : ProcessingSource
+        : ProcessingSource,
+          IDisposable
     {
         private IEnumerable<Option> supportedOptions = new List<Option>()
         {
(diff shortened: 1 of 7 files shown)


src/Microsoft.Performance.Toolkit.Engine.Tests/TestCookers/Source123/Source123DataSource.cs
```diff
@@ -70,5 +80,29 @@ public IEnumerable<TableDescriptor> Discover(ITableConfigurationsSerializer tabl
                 };
             }
         }
+
+        private void Dispose(bool disposing)
+        {
+            if (!disposedValue)
+            {
+                if (disposing)
```
anramanath, line 88: Do we need this if statement and braces?

anramanath: I think we can simplify this method to:

`disposedValue = true`

since we don't seem to care if we've somehow already disposed and the dispose method is being called.

jmaxson-ms: this is boilerplate code... literally generated by VS for us.

helenkzhang (approved):

src/Microsoft.Performance.Toolkit.Engine.Tests/TestCookers/Source123/Source123DataSource.cs
jmaxson-ms: Also, this is a good reminder if this code is ever updated to only dispose of managed objects in the correct location.

The structure of this method *always* should have something similar to this. Managed objects should only be disposed of in the "if (disposing)" code block.

Given all of this, I prefer to keep this as-is.

mslukebo (approved):

PR #349, opened 2024-03-21
Fix FileSystemObsoletePluginsRemover for in-use plugins

mslukebo:
Fixes two issues
1. If a plugin is in use, the move may sometimes still succeed but the delete fail. To prevent modifying a plugin that is loaded, we first verify every single file within the plugin's directory is not currently in use. If anyone currently holds a handle to any file, we abort. Otherwise, we know it's at the very least safe to move, because it's impossible for another instance of a plugins system to return the directory as a plugin to be loaded (since we hold the registry lock).
2. If a move succeeds but delete fails, future attempts to move the directory will fail because the folder exists. To fix this case, the folder moved to now contains a random string.

### DIFF
--- a/src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/Storage/FileSystemObsoletePluginsRemover.cs
+++ b/src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/Storage/FileSystemObsoletePluginsRemover.cs
@@ -161,7 +161,7 @@ namespace Microsoft.Performance.Toolkit.Plugins.Runtime.Installation
                     {
                         return File.Open(f, FileMode.Open, FileAccess.ReadWrite, FileShare.Delete);
                     }
-                    catch (IOException)
+                    catch (Exception)
                     {
                         return null;
                     }
--- a/src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/Storage/FileSystemObsoletePluginsRemover.cs
+++ b/src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/Storage/FileSystemObsoletePluginsRemover.cs
@@ -116,43 +116,52 @@ namespace Microsoft.Performance.Toolkit.Plugins.Runtime.Installation
 
                 /*
                  * The directory is not in use, and the plugins system won't return that it is available
-                 * to be loaded because we currently hold the lock on the plugin registry. We will first
-                 * move the directory to a temporary location, and then delete it. The delete may fail if
-                 * another process opens a file in the directory between the move and the delete, but that's OK
-                 * because we only care that no nobody tries to load the plugin via the plugins system.
+                 * to be loaded because we currently hold the lock on the plugin registry. We already deleted
+                 * all of the files that were originally in the directory, so we just need to clean up the directory
+                 * itself. Note that between deleting all the contents above and now, there could potentially be new
+                 * files that were added (and which are now in use). Whatever those files are and whoever is using them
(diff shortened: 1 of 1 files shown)


src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/Storage/FileSystemObsoletePluginsRemover.cs
```diff
@@ -137,5 +150,35 @@ private Task<bool> TryDeleteDirectory(string dir, CancellationToken cancellation
...
+            List<string> files = Directory.EnumerateFiles(directory, "*", SearchOption.AllDirectories).ToList();
+
+            List<FileStream> streams = files
+                .Select(f =>
+                {
+                    try
+                    {
+                        return File.Open(f, FileMode.Open, FileAccess.ReadWrite, FileShare.Delete);
+                    }
+                    catch (IOException)
```
jmaxson-ms, line 164: should we just catch any exception here?

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/Storage/FileSystemObsoletePluginsRemover.cs
```diff
@@ -107,10 +108,22 @@ private Task<bool> TryDeleteDirectory(string dir, CancellationToken cancellation
...
-                var toDelete = Path.Combine(Path.GetTempPath(), new DirectoryInfo(dir).Name + ".delete");
+                if (!TryDeleteContentsAtomic(dir))
+                {
+                    this.logger.Error($"Failed to delete {dir} because it is in use.");
+                    return false;
+                }
+
+
+                /*
+                 * The directory is not in use, and the plugins system won't return that it is available
```
helenkzhang, line 119: Shouldn't all files under the directory have been already deleted if they can at this point? I wonder if we still need to call `Directory.Move` now that we know the success doesn't guarantee the directory not in use? Can we just delete the current directory?

mslukebo: Not necessarily true. There may have been a new file added to the directory after it was enumerated.

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/Storage/FileSystemObsoletePluginsRemover.cs
mslukebo: Sure

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Runtime/Installation/Storage/FileSystemObsoletePluginsRemover.cs
mslukebo: I updated this code path a bit. Now, first we'll just attempt to delete the directory like you suggest. If that fails, we'll then attempt to rename the directory.

mslukebo: Note that I also changed the move to just rename the folder within the same parent folder, instead of the temp directory. The reasoning here is
1. You can't move a directory across volumes. If the plugin is installed on a separate drive than the temp directory, the move will fail
2. If we move within the same folder, the next time cleanup code runs it'll attempt to delete it again.

helenkzhang (approved):

PR #353, opened 2024-04-11
Add support for static generic methods as bodies for struct functions

crisl-ms:
# Problem

The name of the implicitly created struct function type for a static method is fully determined by full name of the static method declaring type and the method name itself.  This leads to runtime exceptions when multiple instantiations of the same generic method or multiple overloads with the same name are used as bodies for multiple concurrent struct functions -- because of a type name collision within the same code generation context module.

# Solution

We can eliminate this type name collision by adding a unique discriminator between generic instantiations and/or overloads.  Such a unique discriminator is the value of the associated `RuntimeMethodHandle`.

This enables the free use of static generic methods and static method overloads as bodies for struct functions, leading to better code sharing and simpler code free of extraneous ad-hoc discriminators.

### DIFF
--- a/src/Microsoft.Performance.SDK/Processing/CodeGenerationContext.cs
+++ b/src/Microsoft.Performance.SDK/Processing/CodeGenerationContext.cs
@@ -110,7 +110,7 @@ namespace Microsoft.Performance.SDK.Processing
             var resultType = methodToCall.ReturnType;
 
             var type = moduleBuilder.DefineType(
-                methodToCall.DeclaringType.FullName + "$." + methodToCall.Name,
+                $"{methodToCall.DeclaringType.FullName}$.{methodToCall.Name}$.{methodToCall.MethodHandle.Value.ToInt64():x16}",
                 TypeAttributes.Public,
                 typeof(ValueType),
                 new Type[] 
--- a/src/Microsoft.Performance.SDK/Processing/CodeGenerationContext.cs
+++ b/src/Microsoft.Performance.SDK/Processing/CodeGenerationContext.cs
@@ -110,6 +110,20 @@ namespace Microsoft.Performance.SDK.Processing
             var resultType = methodToCall.ReturnType;
 
             var type = moduleBuilder.DefineType(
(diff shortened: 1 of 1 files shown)


src/Microsoft.Performance.SDK/Processing/CodeGenerationContext.cs
```diff
@@ -110,7 +110,7 @@ private static Type CreateStructFunctionType(MethodInfo methodToCall)
             var resultType = methodToCall.ReturnType;
 
             var type = moduleBuilder.DefineType(
-                methodToCall.DeclaringType.FullName + "$." + methodToCall.Name,
+                $"{methodToCall.DeclaringType.FullName}$.{methodToCall.Name}$.{methodToCall.MethodHandle.Value.ToInt64():x16}",
```
jmaxson-ms, line 113: Thanks for this Cristian!

Would you mind adding a comment here to explain that we're using the MethodHandle as a way to uniquely define the type name in cases where there might be a name collision.

It is obvious in context of this PR, but some poor dev looking at this in 3 years might be scratching their head.

jmaxson-ms: Given that plugin assemblies may be loaded and unloaded within a given process, and that the result of this method is stored in a static value which may remain loaded, if using the method handle is safe.

jmaxson-ms: I think this will be ok because it's keyed on the MethodInfo and so the name shouldn't be reused in this case.

crisl-ms: Added comment explaining intent and effectiveness of using `MethodInfo.MethodHandle.Value` in the name of the generated type, including why the `methodToCall` cannot be unloaded while the generated type is still loaded.

jmaxson-ms (approved):

trgibeau (approved):

PR #360, opened 2024-05-17
Add ability to register and retrieve auth providers to/from IApplicationEnvironment

### LINKED ISSUE #359: Add support for specifying authentication providers
**Is your feature request an entirely new concept?**
Yes.

**Is your feature request related to an existing component?**
No.

**Is your feature request related to a problem? Please describe.**
In many situations, a plugin may require performing some kind of authentication (e.g. fetching access tokens on behalf of a user). Currently, a plugin would need to implement all the logic for performing the authentication. This can cause issues where, from the perspective of the user, authentication flows "randomly" happen.

**Describe the solution you'd like**
Authentication flows should happen at the **application** (SDK Driver) layer, **not** at the plugin layer. A plugin should **initiate** these flows, but the logic for how they happen should be defined by the application. Specifically, the application environment should provide ways to fetch classes that can provide authentication flows. Engine users should also be able to provide these classes, which then are added to the engine application environment.

mslukebo:
Closes #359

### DIFF
--- a/src/Microsoft.Performance.SDK.Runtime/ApplicationEnvironment.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/ApplicationEnvironment.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using Microsoft.Performance.SDK.Auth;
 using Microsoft.Performance.SDK.Extensibility.DataCooking;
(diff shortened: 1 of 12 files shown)


trgibeau (approved):

jmaxson-ms (approved):
:shipit:

PR #363, opened 2024-07-09
Doc Suggestions

karthik-sesham:
Fixed some small typos and added some more information that could be helpful

### DIFF
--- a/documentation/Using-the-SDK/Using-the-engine.md
+++ b/documentation/Using-the-SDK/Using-the-engine.md
@@ -10,6 +10,9 @@ major components to be aware of when using the `Engine`:
 2. The [`DataSourceSet`](#DataSourceSet)
 3. The [`Engine`](#Engine) itself
 
+Before starting, make sure to include the `Microsoft.Performance.Toolkit.Engine`
+library to use the engine.
+
 ## Plugin Set
 
 The `PluginSet` is a collection of all of your plugins and extensions that can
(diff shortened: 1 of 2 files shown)


documentation/Using-the-SDK/Using-the-engine.md
```diff
@@ -118,6 +121,26 @@ using (var dataSources = DataSourceSet.Create(plugins))
 }
 ````
 
+If you would like to enable a composite cooker, call the `GetCookedData(DataCookerPath)`
```
mslukebo, line 124: It's slightly incorrect to say you need to do this to "enable" a composite cooker. The composite cooker is enabled the same way any other cooker is enabled. I think instead we could have a "NOTE" here. Something like

```md
> ⚠️ Note: [Composite cookers](../Glossary.md#compositedatacooker) are processed lazily. The above code will not cause any code inside of an enabled composite cooker to execute. To execute/debug composite cookers, you must also have code that queries for the specific cooker, such as building a table that uses the cooker or manually calling `results.GetCookedData(CompCookerPath)` after processing.
```

Which would look like

> ⚠️ Note: [Composite cookers](../Glossary.md#compositedatacooker) are processed lazily. The above code will not cause any code inside of an enabled composite cooker to execute. To execute/debug composite cookers, you must also have code that queries for the specific cooker, such as building a table that uses the cooker or manually calling `results.GetCookedData(CompCookerPath)` after processing.

mslukebo:
Closing in favor of #372 which uses the above change.

PR #365, opened 2024-07-16
Move TryGetAuthProvider to new IApplicationEnvironmentV2 interface

mslukebo:
This ensures backwards compatibility with plugins on 1.x

### DIFF
--- a/src/Microsoft.Performance.SDK/Processing/IApplicationEnvironmentV2.cs
+++ b/src/Microsoft.Performance.SDK/Processing/IApplicationEnvironmentV2.cs
@@ -3,6 +3,9 @@ using Microsoft.Performance.SDK.Auth;
 
 namespace Microsoft.Performance.SDK.Processing
 {
+    /// <summary>
+    ///     Extends <see cref="IApplicationEnvironment"/> to provide additional functionality.
+    /// </summary>
     [Obsolete("This interface will be removed in version 2.0 of the SDK. It is OK to use this interface in version 1.x of the SDK.")]
     public interface IApplicationEnvironmentV2
         : IApplicationEnvironment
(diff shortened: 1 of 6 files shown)


trgibeau (approved):

src/Microsoft.Performance.SDK/Processing/IApplicationEnvironmentV2.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using Microsoft.Performance.SDK.Auth;
+
+namespace Microsoft.Performance.SDK.Processing
+{
+    /// <summary>
+    ///     Extends <see cref="IApplicationEnvironment"/> to provide additional functionality.
+    /// </summary>
+    [Obsolete("This interface will be removed in version 2.0 of the SDK. It is OK to use this interface in version 1.x of the SDK.")]
+    public interface IApplicationEnvironmentV2
```
jmaxson-ms, line 10: >IApplicationEnvironmentV2 [](http://example.com/codeflow?start=21&length=25)

are we using a 'V' for version in interfaces? (instead of 'IApplicationEnvironment2')?

jmaxson-ms, line 10: are we using a 'V' for version in interfaces? (instead of 'IApplicationEnvironment2')?

jmaxson-ms: not sure why this was duplicated

jmaxson-ms: I wonder if this shouldn't extend IApplicationEnvironment, but should just be a different interface that is implemented by ApplicationEnvironment.

mslukebo: That wouldn't work because plugins only receive `IApplicationEnvironment` instance (not concrete `ApplicationEnvironment` instances) and we need a way for them to call this method. Currently that's with the IAppEnv extension method this PR adds.

jmaxson-ms (approved):

PR #368, opened 2024-09-25
Allow per processor table synchronization.

jmaxson-ms:
Table synchronization at the processor level provides greater control over which columns are required to be updated when applying symbols changes.

Attempting to preserve backcompat by adding new, optional interfaces.


trgibeau (approved):

trgibeau (approved):

src/Microsoft.Performance.SDK.Runtime/ApplicationEnvironment.cs
```diff
@@ -148,5 +149,29 @@ private static MessageBoxIcon MapToImage(MessageType messageType)
...
+        private sealed class NullTableSynchronizer
+            : ITableDataSynchronization
+        {
+            public void SubmitColumnChangeRequest(
+                IEnumerable<Guid> columns,
+                Action onReadyForChange,
+                Action onChangeComplete,
+                bool requestInitialFilterReevaluation = false)
+            {
+                onReadyForChange?.Invoke();
```
mslukebo, line 162: according to the spec of this interface, these calls should happen in a lock

jmaxson-ms: Haha, I copied and pasted this from the runtime implementation.

jmaxson-ms: Actually, we shouldn't need the runtime implementation anymore given this change.

mslukebo: Perhaps we should add the lock anyway :)

jmaxson-ms: Do you mean this comment?

`Lock tables affected by the change (based on <paramref name="predicate"/>) and wait for all existing operations on those tables to complete.`

mslukebo (approved):

trgibeau (approved):

mslukebo (approved):

Issue #237, opened 2023-01-23
Unable to set "Custom Format" of timestamp columns


wmmc88:
**Describe the bug**
In the WPA UI, columns showing timestamp data can adjust the unit of time used via the "Custom Format" option:
![image](https://user-images.githubusercontent.com/16629657/214116429-dbc9730d-d234-46f5-91c8-e519b55b6f92.png)

When construction a table using this SDK, the UI hints does not seem to have any options to set the Custom Format.

**Expected behavior**
API should expose the ability to set time units of timestamp column.

**Version information (please complete the following information):**
 - SDK version: 1.0.16
 - WPA version (if applicable): 11.1.5.2

wmmc88:
possibly related to #208?

helenkzhang:
@wmmc88 Thanks for your feedback. Unfortunately, this is not currently supported by the SDK. 
And you're right, this is tracked by issue https://github.com/microsoft/microsoft-performance-toolkit-sdk/issues/208 and will be added as a new SDK feature.

Issue #208, opened 2022-09-07
SDK should support Specialized Column


helenkzhang:
## Overview
The SDK currently defines a column as a conceptual pair of `Column Configuration` and `Projection`. When a column is added to a table builder, it is associated with one single projection and one single configuration. However, there are use cases where a column is expected to have special modes – different ways of projecting and displaying the column data – that can be specified and changed at runtime. For example, a process column should be able to display process name only, PID only or both process name and PID.  Currently the functionality of extending a column can only be implemented in the SDK driver but not in the plugin because of the lack of support from the SDK. This hinders plugin authors from creating their own column extensions and introduces unwanted plugin-specific code into the SDK driver. To address this, we want to introduce the concept of `Specialized Column` and add support for creating these columns using the SDK.

#### Build a specialized column
- Create a **_column extender_** that provides APIs needed to create a specific column variation and stores a _**column extension descriptor**_ as the key to identify an extension mode
- Create a _**specialized column**_ (regular or hierarchical) that stores a collection of supported _**column extenders**_
#### Add a specialized column to a table
- Add a specialized column to a table
- Add new extenders to an already created specialized column
#### Configure a specialized column
- Define a set of customized configurable _**column extension properties**_ associated with a specialized column (e.g. a boolean that specifies whether a specialized hierarchical column should be fold). These properties will be used by the extender to build projection and potentially other column properties. To avoid arbitrary types, we can define an enum of supported "types" of configurable properties and have the properties be of type `ExtensionProperty` which itself has a property of this enum
- Create an **_extended column configuration_** that specifies an extension mode and the properties values associated with the specialized column
### Code sample with new types and APIs
- Note that this is not finished code but just to illustrate one design idea.  [Early ideas for adding extensible column support · microsoft/microsoft-performance-toolkit-sdk@3d03660 (github.com)](https://github.com/microsoft/microsoft-performance-toolkit-sdk/compare/user/helzhang/extensibleColumn)

PR #370, opened 2024-10-16
Add support for specifying variants of added columns during table building

### LINKED ISSUE #373: Add APIs for building + fetching column variants

mslukebo:
# Overview

This PR exposes three APIs that together address #208 and #237, and closes #373: 
1. Add APIs in the runtime for plugins to specify _column modes_ during table building.
2. Add APIs in the runtime for SDK drivers to discover and fetch modes registered for individual columns.
3. Add APIs in the Engine for fetching registered modes

A _column mode_ is defined as an alternative `IProjection<int, T>` for a column. When adding column modes during table building, plugins supply
1. A `ColumnVariantIdentifier` that contains a `Guid` uniquely identifying the added variant
2. An `IProjection<int, T>` to use for that variant. Note that variant projections need not return the same type as the base column's type.

Registered column variants are exposed to consumers as an untyped `IDataColumn` whose `ColumnConfiguration` is the same as the configuration of the variant's base column. 

Column variants also have meta-information about how the variants relate to the base column and each other. For example, a base column of type `IDataColumn<Timestamp>` may have a variant projection `IDataColumn<DateTime>` that exposes the same time value as a `DateTime` time relative to some fixed point in time. The meta-information is that the base column can be **toggled** to _"project timestamp as DateTime"_.

Continuing with this example, suppose the base `Timestamp` column can project to different `DateTime`s that are relative to different points in time (e.g. one `DateTime` relative to the start of the trace and one `DateTime` relative to the first event). In this scenario, the meta-information is that the base column can be **toggled** to "project timestamp as DateTime" in **one of two modes**.

This meta-information is defined during table building through the API exposed to plugins. Variants are registered through `ColumnBuilder` classes, where both the methods used and the order that they are used dictate variants' relationships. For concrete examples of how this looks, refer to `ColumnVariantsTests.cs`

Meta-information along with concrete column variants are exposed as a tree structure accessible through a new `IColumnVariantsRegistrar` interface. This interface is exposed on the runtime's `TableBuilder` class (but not the `ITableBuilder` interface passed to plugins). The `IColumnVariantsRegistrar` allows SDK drivers to fetch `IColumnVariantsTreeNode` instances that represent the root of registered variants. Drivers can implement a `IColumnVariantsTreeNodesVisitor` that can traverse the tree data structure to understand both
1. The meta-information defined during table building 
2. The concrete variants associated with different nodes in the tree, exposed as `IDataColumn`s on tree nodes

Finally, the Engine's `ITableResult` was updated to expose all possible `ColumnVariantIdentifier`s for `IDataColumn`s on the table. This representation loses the meta-information that defines how the variants relate to each other, which is OK since the Engine is used for programmatic access to data where the conceptual relationship between variants is not necessary.

# Column Builders API Design
To be consist with the "builder" architecture used to construct tables, plugins use a builder pattern to add variants of a column. The following methods were added:
1. `ITableBuilderWithRowCount.AddColumnWithVariants`
2. `TableBuilderExtensions.AddColumnWithVariants`
3. `TableBuilderExtensions.AddHierarchicalColumnWithVariants`

These methods follow the same pattern of `AddColumn` and `AddHierarchicalColumn`, where the `ITableBuilderWithRowCount` method takes in the `IDataColumn` being added and the `TableBuilderExtensions` methods take in the column configuration + projection. These three methods have an additional parameter `Action<IRootColumnBuilder> variantsBuilder` that allow plugins to provide a callback that configures the added column's variants. A plugin's code might look like

```cs
tableBuilder
    .SetRowCount(1)
    .AddColumnWithVariants(baseConfig, baseProj, builder =>
    {
        builder
            .WithToggle(projectAsDateTime, utcProj)
            .Build();
    });
```
There are two differences between the column builders' design and the table builders' design:
1. Column builders are purely functional with no mutable state. Every method on the builder returns an entirely new object
2. Plugins **must** call `Build()` on the column builder for any change to have an effect

This decision was made because, unlike table builders, both the final state and valid operations of a column builder depends on the order methods were called. For example, once a base column has modes added, it should not be able to add a toggle at the same level of the group of modes (adding toggles beneath an individual mode, however, is valid). I.e., this must be invalid:

```cs
tableBuilder
    .SetRowCount(1)
    .AddColumnWithVariants(baseConfig, utcProj, builder =>
    {
        builder
            .WithModes(utc.Name)
            .WithMode(local, localProj)
            .WithToggle(projectAsDateTime, utcProj) // INVALID CALL!
            .Build();
    });
```

Ensuring `WithToggle` cannot be called after `WithModes` is added is accomplished by the return type of `WithModes` being a different interface than `IRootColumnBuilder`. Specifically, `WithModes` returns an `IModalColumnBuilder` that does not expose a `WithToggle` method.

However, this is not enough to ensure invalid states are impossible. If the column builder was a single, stateful object, the plugin could write this code:

```cs
tableBuilder
    .SetRowCount(1)
    .AddColumnWithVariants(baseConfig, utcProj, builder =>
    {
        builder
            .WithModes(utc.Name)
            .WithMode(local, localProj);

        builder
            .WithToggle(projectAsDateTime, utcProj) // INVALID CALL!
            .Build();
    });
```

If `builder` was not a functional object, it would be forced to throw an exception on the invalid call. Instead, the behavior of the above code is to completely disregard the added modes since the object that was actually built was just a single toggle. The built column will therefore have just a single toggle underneath the base column.

# Column Builders Implementation
There are several classes that implement the various `ColumnBuilder` abstract classes depending on what has been added. For example, there is one implementation of an `RootColumnBuilder` that has had nothing added to it.

This PR adds an `IColumnVariantsProcessor` interface to the runtime. This interface is passed to `ColumnBuilder` implementations to be invoked during their `Build` calls. When a builder is built, it constructs a column variants tree based on that builder's data and then passes it to the processor it was given.

The runtime also defines a helper interface `IBuilderCallbackInvoker` that lets column builders fetch an `IColumnVariantsTreeNode` for children of tree nodes it is building. The implementations of this interface run the plugin-supplied `Action<T> where T : ColumnBuilder` callbacks, returning the final root tree node built by the callback.


src/Microsoft.Performance.Toolkit.Engine/ITableResult.cs
```diff
@@ -28,6 +29,14 @@ public interface ITableResult
         /// </summary>
         IReadOnlyCollection<IDataColumn> Columns { get; }
 
+        /// <summary>
+        ///     Gets the mapping of columns that have been added to the table with variants to their added variants.
```
jmaxson-ms, line 33: >Gets the mapping of columns that have been added to the table with variants to their added variants. [](http://example.com/codeflow?start=16&length=100)

this wording is a bit confusing

src/Microsoft.Performance.Toolkit.Engine/ITableResultExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+        public static bool TryGetColumnVariants(
+            this ITableResult self,
+            IDataColumn baseColumn,
+            out IReadOnlyDictionary<ColumnVariantIdentifier, IDataColumn> foundColumns)
+        {
+            return self.ColumnVariants.TryGetValue(baseColumn, out foundColumns);
+        }
+
+        /// <summary>
+        ///     Attempts to get the column variant for the given column and identifier.
```
jmaxson-ms, line 41: >identifier [](http://example.com/codeflow?start=76&length=10)

'variant identifier' might make this a bit more clear

src/Microsoft.Performance.Toolkit.Engine/ITableResultExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            if (!self.TryGetColumnVariants(column, out var variants))
+            {
+                columnVariant = null;
+                return false;
+            }
+
+            return variants
+                .ToDictionary(kvp => kvp.Key.Guid, kvp => kvp.Value)
+                .TryGetValue(variantGuid, out columnVariant);
```
jmaxson-ms, line 79: This feels like it might not be especially performant and will memory allocation. Might not be a big deal, depending on how it's used, but we might be able to make this more efficient.

src/Microsoft.Performance.Toolkit.Engine/ITableResult.cs
mslukebo: Updating to `Gets the mapping from columns with variants to those variants.`

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnVariantIdentifier.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Performance.SDK.Processing.ColumnBuilding;
+
+/// <summary>
+///     Represents a identifier for a column variant.
```
jmaxson-ms, line 9: >a [](http://example.com/codeflow?start=19&length=1)

nit: an identifier

jmaxson-ms, line 9: >column variant [](http://example.com/codeflow?start=38&length=14)

Is a column variant defined in a doc comment somewhere? It seems like something we should define as we use it quite often. If not, would this be a decent place to do so?

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/IColumnBuilder.cs
```diff
@@ -0,0 +1,18 @@
...
+///     part of a table.
+/// </summary>
+public interface IColumnBuilder
+{
+    /// <summary>
+    ///     Builds the column. Note that this method MUST be called to have
+    ///     any effect on the column being built. Calling this method multiple
+    ///     times will result in only the final call having any effect.
+    /// </summary>
+    void Build();
```
jmaxson-ms, line 17: >Build(); [](http://example.com/codeflow?start=9&length=8)

Opinion:

When I first read through ColumnVariantsTests.BuildingTwiceOverridesFirstBuild(), I was scratching my head trying to understand what was going on.

I believe the issue for me is the naming of this method is non-intuitive.

Sure, it builds the variant, but that part is mostly meaningless because, if it's not *applied* to the processor, then it does nothing - and that's the key to this method.

I believe that changing the name to Apply() or Process() or similar would make using this more intuitive.

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/IColumnBuilder.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Performance.SDK.Processing.ColumnBuilding;
+
+/// <summary>
+///     Base interface for classes that can build/extend upon columns that are
+///     part of a table.
+/// </summary>
+public interface IColumnBuilder
```
jmaxson-ms, line 10: >IColumnBuilder [](http://example.com/codeflow?start=17&length=14)

Should we make this name more specific, maybe by adding Variant?

mslukebo: I would prefer not to in case we ever want to expose more operations on columns

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnVariantIdentifier.cs
```diff
@@ -0,0 +1,17 @@
...
+/// </summary>
+/// <param name="Guid">
+///     The unique identifier for the column variant.
+/// </param>
+/// <param name="Name">
+///     The human-readable name of the column variant.
+/// </param>
+public record ColumnVariantIdentifier(Guid Guid, string Name);
```
jmaxson-ms, line 17: Would it be helpful to add a description to this? I'm thinking the description might be helpful in a tooltip or similar.

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/IColumnBuilder.cs
mslukebo: I can see that confusion. `Build` usually returns an object being built.

I am OK with `Apply()`. However, I think `Commit()` might also work. Do you prefer one over the other?

jmaxson-ms: Hrm, I think that I like Commit

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/IModalColumnBuilder.cs
```diff
@@ -0,0 +1,151 @@
...
+public interface IModalColumnBuilder
+    : IColumnBuilder
+{
+    /// <summary>
+    ///     Adds a mode to the column.
+    /// </summary>
+    /// <param name="modeIdentifier">
+    ///     The <see cref="ColumnVariantIdentifier"/> for the mode. The
+    ///     <see cref="ColumnVariantIdentifier.Name"/> represents the name of
+    ///     this mode.
```
jmaxson-ms, line 21: nit: I don't think this second sentence adds much. Why call out the name specifically?

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnVariantIdentifier.cs
jmaxson-ms: If we do decide to add a description, perhaps we should change the name from Identifier to Descriptor, or something similar.

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/IModalColumnBuilder.cs
```diff
@@ -0,0 +1,151 @@
...
+        ColumnVariantIdentifier modeIdentifier,
+        IProjection<int, T> projection);
+
+    /// <summary>
+    ///     Adds a hierarchical mode to the column.
+    /// </summary>
+    /// <param name="modeIdentifier">
+    ///     The <see cref="ColumnVariantIdentifier"/> for the mode. The
+    ///     <see cref="ColumnVariantIdentifier.Name"/> represents the name of
+    ///     this mode.
```
jmaxson-ms, line 46: Same as above.

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnVariantIdentifier.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System;
+
+namespace Microsoft.Performance.SDK.Processing.ColumnBuilding;
+
+/// <summary>
+///     Represents a identifier for a column variant.
+/// </summary>
+/// <param name="Guid">
+///     The unique identifier for the column variant.
```
jmaxson-ms, line 12: >The unique identifier for the column variant. [](http://example.com/codeflow?start=8&length=45)

Does this only need to be unique within a column - meaning it can be used in other columns without an issue?

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnVariantIdentifier.cs
jmaxson-ms, line 17: >Guid Guid [](http://example.com/codeflow?start=38&length=9)

Do we have any checks on uniqueness? Do we want any? Seems like it might be helpful to catch accidental copy/paste errors early as that might be a pain to debug.

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnVariantIdentifier.cs
mslukebo: I think by definition these identifiers need to be **globally** unique.

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnVariantIdentifier.cs
mslukebo: I can add some checks. There are two options
1. Verify GUIDs as they are added with each method call
2. Check all GUIDs during the root's `Commit` (`Build`) call

Option 1 would provide more immediate feedback, but would require a walk of the entire data structure for every method call. It's also harder to implement.

Option 2 would be easier to implement, but provide less immediate feedback.

mslukebo: Added option 2 (with a unit test)

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/IColumnBuilder.cs
mslukebo: Changed to Commit

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnVariantIdentifier.cs
mslukebo: I like the idea of using descriptor, since technically the Guid is what actually identifies the variant.

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnVariantIdentifier.cs
mslukebo: Added

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnBuilder.cs
```diff
@@ -0,0 +1,26 @@
...
+/// </summary>
+public abstract class ColumnBuilder
+{
+    internal ColumnBuilder()
+    {
+    }
+
+    /// <summary>
+    ///     Commits this builder's configuration to the column being built. Note that this method MUST be called to have
+    ///     any effect on the column being built. Calling this method multiple times will override previous calls.
```
jmaxson-ms, line 20: > Calling this method multiple times will override previous calls. [](http://example.com/codeflow?start=49&length=65)

I've been thinking about the flexibility of being able to call this more than once. Should we allow it? Would it better to throw if this has already been called?

It seems like the pattern we'd expect it to use the builder to build and build until the final tree is ready, then commit once.

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ModalColumnBuilder.cs
```diff
@@ -0,0 +1,155 @@
...
+namespace Microsoft.Performance.SDK.Processing.ColumnBuilding;
+
+/// <summary>
+///     A builder for configuring modes on a column that has been
+///     configured as having mutually exclusive modes.
+/// </summary>
+public abstract class ModalColumnBuilder
+    : ColumnBuilder
+{
+    internal ModalColumnBuilder()
```
jmaxson-ms, line 15: >internal [](http://example.com/codeflow?start=4&length=8)

same here and for abstract classes in general: some sort of protected constructor - private protected in this case

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnBuilder.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Microsoft.Performance.SDK.Processing.ColumnBuilding;
+
+/// <summary>
+///     Base class for classes that can build/extend upon columns that are
+///     part of a table.
+/// </summary>
+public abstract class ColumnBuilder
+{
+    internal ColumnBuilder()
```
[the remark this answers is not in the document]
mslukebo: Thanks, for about that one

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnVariantIdentifier.cs
trgibeau: hmm, what do we do for scenarios where a column projection is reused across multiple tables?

mslukebo: The projection itself can be re-used, but if the same projection is being used for multiple columns' variants they need different GUIDs.

mslukebo: Hmm, thinking about that scenario more, I can see a benefit of letting different columns have the same GUIDs for similar variants (e.g. the "project as datetime" variant).

I think it's OK to duplicate those GUIDs. We should only really be searching for a guid per-column anyway. I'll update the docs.

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnBuilder.cs
mslukebo: After discussing this offline, I have changed this API to
1. Make this method `internal`
2. Changing column building APIs to take in functions that return the final column builders they want to apply
3. Invoking Commit on the returned builders from within the runtime/engine

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ModalColumnBuilder.cs
```diff
@@ -0,0 +1,149 @@
...
+    ///     Adds a mode to the column.
+    /// </summary>
+    /// <param name="modeDescriptor">
+    ///     The <see cref="ColumnVariantDescriptor"/> for the mode.
+    /// </param>
+    /// <param name="projection">
+    ///     The projection that will be used to generate the column for this mode.
+    /// </param>
+    /// <param name="builder">
+    ///     A callback that builds sub-variants of the added mode and returns its final column configuration..
```
jmaxson-ms, line 80: >. [](http://example.com/codeflow?start=109&length=1)

nit: extra period

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ModalColumnBuilder.cs
```diff
@@ -0,0 +1,149 @@
...
+    /// <returns>
+    ///     A new instance of <see cref="ColumnBuilder"/> that has been
+    ///     configured with the default mode.
+    /// </returns>
+    /// <exception cref="ArgumentException">
+    ///     The mode with the given <paramref name="modeIdentifierGuid"/> has not been
+    ///     added as an available mode.
+    /// </exception>
+    /// <remarks>
+    ///     If this method is not called prior to <see cref="ColumnBuilder.Commit"/>,
```
jmaxson-ms, line 145: ><see cref="ColumnBuilder.Commit"/>, [](http://example.com/codeflow?start=50&length=35)

we probably don't want to reference Commit here anymore

src/Microsoft.Performance.SDK.Runtime/ColumnBuilding/Builders/CallbackInvokers/ModesBuilderCallbackInvoker.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        var processor = new BuiltColumnVariantReflector();
+        var builder = new ModalColumnWithModesBuilder(
+            processor,
+            new List<ModalColumnWithModesBuilder.AddedMode>(),
+            this.baseColumn,
+            null);
+
+        this.callback(builder).Commit();
+        builtVariantsTreeNode = processor.Output;
```
trgibeau, line 55: nit: drop this.

trgibeau: or add this. to the callback above, and in the other callback class

src/Microsoft.Performance.SDK.Runtime/ColumnBuilding/Builders/CallbackInvokers/ModeBuilderCallbackInvoker.cs
```diff
@@ -0,0 +1,53 @@
...
+using Microsoft.Performance.SDK.Processing.ColumnBuilding;
+using Microsoft.Performance.SDK.Runtime.ColumnBuilding.Processors;
+using Microsoft.Performance.SDK.Runtime.ColumnVariants.TreeNodes;
+
+namespace Microsoft.Performance.SDK.Runtime.ColumnBuilding.Builders.CallbackInvokers;
+
+/// <summary>
+///     Responsible for invoking a callback that builds a single mode of a modal column variant.
+/// </summary>
+internal sealed class ModeBuilderCallbackInvoker
```
jmaxson-ms, line 15: nit: these implementation are so small, could we just make them structs? not a big deal at all if you don't want to, but seems like they're ideal candidates.

src/Microsoft.Performance.SDK.Runtime/ColumnBuilding/Processors/BuiltColumnVariantsRegistrant.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.Performance.SDK.Runtime.ColumnVariants.Registrar;
+using Microsoft.Performance.SDK.Runtime.ColumnVariants.TreeNodes;
+using Microsoft.Performance.SDK.Runtime.ColumnVariants.TreeNodes.Visitors;
+
+namespace Microsoft.Performance.SDK.Runtime.ColumnBuilding.Processors;
+
+/// <summary>
+///     Responsible for registering built column variants with a table builder.
+/// </summary>
+public class BuiltColumnVariantsRegistrant
```
jmaxson-ms, line 16: can we `seal` this class?

src/Microsoft.Performance.SDK.Runtime/ColumnBuilding/Processors/BuiltColumnVariantReflector.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using Microsoft.Performance.SDK.Runtime.ColumnVariants.TreeNodes;
+
+namespace Microsoft.Performance.SDK.Runtime.ColumnBuilding.Processors;
+
+/// <summary>
+///     Exposes the <see cref="IColumnVariantsTreeNode"/> that was passed to the processor.
+/// </summary>
+internal class BuiltColumnVariantReflector
+    : IColumnVariantsProcessor
```
jmaxson-ms, line 12: could we either `sealed` or struct?

mslukebo: Made this sealed class since making it a struct breaks fetching the output since this is passed to the builders

PR #379, opened 2024-10-23
Add support for specifying column variants in a ColumnConfiguration

### LINKED ISSUE #375: Support column variant identifiers in table configurations

mslukebo:
Closes #375 

To prevent major breaking changes to the `TableConfiguration` API, this PR just adds a new property off the `ColumnConfiguration` for this information. New DTO objects were created to facilitate serializing this information.

The `ISupportUpgrade` logic was changed to
1. Have earlier versions upgrade to newer versions incrementally, instead of every old version immediately upgrading to the latest. This results in changing O(1) classes instead of O(n) classes when new versions are created. I.e. you only need to add `(n -1) => n` upgrade logic instead of `(n - k) => n` for `k` in `[0, n)`
2. Updating the serializer to recursively upgrade the deserialized version to the latest

For SDK 2.0, the table configuration API can be reworked more aggressively.

### DIFF
--- a/src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
@@ -105,17 +105,11 @@ namespace Microsoft.Performance.SDK.Runtime.DTO
                 // Ensure we don't get stuck in a loop by tracking visited configurations
                 HashSet<PrebuiltConfigurationsBase> visited = new HashSet<PrebuiltConfigurationsBase>();
 
-                while (visited.Add(configurationsBase))
+                while (configurationsBase is not PrebuiltConfigurations &&
+                       configurationsBase is ISupportUpgrade<PrebuiltConfigurationsBase> upgradeable &&
+                       visited.Add(configurationsBase))
                 {
-                    if (configurationsBase is PrebuiltConfigurations)
-                    {
-                        break;
-                    }
-
-                    if (configurationsBase is ISupportUpgrade<PrebuiltConfigurationsBase> upgradeable)
-                    {
-                        configurationsBase = upgradeable.Upgrade();
-                    }
+                    configurationsBase = upgradeable.Upgrade();
                 }
 
                 prebuiltConfigurations = configurationsBase as PrebuiltConfigurations;
--- a/src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
@@ -9,8 +9,8 @@ using System.Runtime.Serialization;
 using System.Runtime.Serialization.Json;
 using System.Text;
 using Microsoft.Performance.SDK.Processing;
-using Microsoft.Performance.SDK.Runtime.DTO.V1_3;
-using TableConfigurations = Microsoft.Performance.SDK.Runtime.DTO.V1_3.TableConfigurations;
+using Microsoft.Performance.SDK.Runtime.DTO.Latest;
+using TableConfigurations = Microsoft.Performance.SDK.Runtime.DTO.Latest.TableConfigurations;
 
 namespace Microsoft.Performance.SDK.Runtime.DTO
 {
@@ -75,7 +75,7 @@ namespace Microsoft.Performance.SDK.Runtime.DTO
                 startingPosition = stream.Position;
             }
 
-            V1_3.PrebuiltConfigurations prebuiltConfigurations = null;
+            PrebuiltConfigurations prebuiltConfigurations = null;
 
             try
             {
@@ -107,7 +107,7 @@ namespace Microsoft.Performance.SDK.Runtime.DTO
 
                 while (visited.Add(configurationsBase))
                 {
-                    if (configurationsBase is V1_3.PrebuiltConfigurations)
+                    if (configurationsBase is PrebuiltConfigurations)
                     {
                         break;
                     }
@@ -118,7 +118,7 @@ namespace Microsoft.Performance.SDK.Runtime.DTO
                     }
                 }
 
-                prebuiltConfigurations = configurationsBase as V1_3.PrebuiltConfigurations;
+                prebuiltConfigurations = configurationsBase as PrebuiltConfigurations;
 
                 if (prebuiltConfigurations == null)
                 {
--- a/src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
@@ -102,12 +102,8 @@ namespace Microsoft.Performance.SDK.Runtime.DTO
                     return Enumerable.Empty<Processing.TableConfigurations>();
                 }
 
-                // Ensure we don't get stuck in a loop by tracking visited configurations
-                HashSet<PrebuiltConfigurationsBase> visited = new HashSet<PrebuiltConfigurationsBase>();
-
                 while (configurationsBase is not PrebuiltConfigurations &&
-                       configurationsBase is ISupportUpgrade<PrebuiltConfigurationsBase> upgradeable &&
-                       visited.Add(configurationsBase))
+                       configurationsBase is ISupportUpgrade<PrebuiltConfigurationsBase> upgradeable)
                 {
                     configurationsBase = upgradeable.Upgrade();
                 }
--- a/src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
@@ -9,6 +9,8 @@ using System.Runtime.Serialization;
 using System.Runtime.Serialization.Json;
 using System.Text;
 using Microsoft.Performance.SDK.Processing;
+using Microsoft.Performance.SDK.Runtime.DTO.V1_3;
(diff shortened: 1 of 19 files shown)


src/Microsoft.Performance.SDK/Processing/ColumnConfiguration.cs
```diff
@@ -39,10 +41,34 @@ public ColumnConfiguration(ColumnMetadata metadata)
...
+        ///     may be <c>null</c>. If this parameter is <c>null</c>, then
+        ///     <see cref="UIHints.Default"/> will be used for this instance.
+        /// </param>
+        /// <param name="variantGuid">
+        ///     Optional unique identifier of the column variant to use.
+        /// </param>
+        /// <exception cref="System.ArgumentNullException">
+        ///     <paramref name="metadata"/> is <c>null</c>.
+        /// </exception>
+        public ColumnConfiguration(ColumnMetadata metadata, UIHints hints, Guid? variantGuid)
```
xj-ms, line 66: Consolidate the ctors with optional parameters?

src/Microsoft.Performance.SDK/Processing/ColumnConfiguration.cs
```diff
@@ -61,6 +87,7 @@ public ColumnConfiguration(ColumnConfiguration other)
             Guard.NotNull(other, nameof(other));
```
xj-ms, line 87: Make this ctor use the one above?

src/Microsoft.Performance.SDK/Processing/ColumnConfiguration.cs
```diff
@@ -85,7 +118,14 @@ public ColumnConfiguration(ColumnConfiguration other)
...
         public override string ToString()
         {
-            return $"{this.Metadata.Guid} - {this.Metadata.Name}";
+            if (this.VariantGuid == null)
+            {
+                return $"{this.Metadata.Guid} - {this.Metadata.Name}";
+            }
+            else
+            {
+                return $"{this.Metadata.Guid} - {this.Metadata.Name} (variant {this.VariantGuid}";
```
xj-ms, line 127: Closing parenthesis?

src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
```diff
@@ -94,15 +96,32 @@ public class TableConfigurationsSerializer
...
                 }
-                else if (desrailizedObject is PrebuiltConfigurations latestConfigs)
+
+                // Ensure we don't get stuck in a loop by tracking visited configurations
+                HashSet<PrebuiltConfigurationsBase> visited = new HashSet<PrebuiltConfigurationsBase>();
+
+                while (visited.Add(configurationsBase))
                 {
-                    prebuiltConfigurations = latestConfigs;
+                    if (configurationsBase is ISupportUpgrade<V1_0.PrebuiltConfigurations> v1_0)
```
xj-ms, line 109: Is it necessary to distinguish the actual version it supports? How about 
```C#
interface UpgradableConfiguration {
  ConfigurationBase Upgrade();
}
```

src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
```diff
@@ -94,15 +96,32 @@ public class TableConfigurationsSerializer
...
+                    {
+                        configurationsBase = v1_0.Upgrade();
+                    }
+                    else if (configurationsBase is ISupportUpgrade<V1_3.PrebuiltConfigurations> v1_3)
+                    {
+                        configurationsBase = v1_3.Upgrade();
+                    }
+                    else if (configurationsBase is V1_3.PrebuiltConfigurations latestConfigs)
```
xj-ms, line 117: Can move this check into the while condition. Also can convert to a `do while`

src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
```diff
@@ -94,15 +96,32 @@ public class TableConfigurationsSerializer
...
-                if (desrailizedObject is ISupportUpgrade<PrebuiltConfigurations> supportUpgrade)
+                if (desrailizedObject is not PrebuiltConfigurationsBase configurationsBase)
                 {
-                    prebuiltConfigurations = supportUpgrade.Upgrade();
+                    return Enumerable.Empty<Processing.TableConfigurations>();
                 }
-                else if (desrailizedObject is PrebuiltConfigurations latestConfigs)
+
+                // Ensure we don't get stuck in a loop by tracking visited configurations
+                HashSet<PrebuiltConfigurationsBase> visited = new HashSet<PrebuiltConfigurationsBase>();
```
xj-ms, line 105: When will there be a cycle? If a configuration wants to upgrade to itself or downgrade then it shouldn't implement `ISupportUpgrade`?

src/Microsoft.Performance.SDK.Runtime/DTO/V1_3/ColumnConfiguration.cs
xj-ms, line 1: If we name the namespace `Latest` instead of `V1_3` we can avoid a lot of renaming next time we have breaking changes

xj-ms (changes requested):

src/Microsoft.Performance.SDK/Processing/ColumnConfiguration.cs
mslukebo: Unfortunately I cannot add or modify the previous constructors because we need to maintain backwards compatibility

xj-ms: I believe having optional parameters still makes it backward compatible. They can still call it with 2 parameters and the third parameter will use the default value

src/Microsoft.Performance.SDK/Processing/ColumnConfiguration.cs
mslukebo: good catch!

src/Microsoft.Performance.SDK/Processing/ColumnConfiguration.cs
mslukebo: Done!

src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
mslukebo: Other classes implement `ISupportUpgrade` where `T` is relevant. However, I made an abstract class that will help with this logic and do what you're suggesting.

src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
mslukebo: It would happen if someone implements the upgrading wrong. I.e. you upgrade from 1.0 to 2.0 then 2.0 "upgrades" to 1.0.

This is mostly to prevent an infinite loop below

src/Microsoft.Performance.SDK.Runtime/DTO/V1_3/ColumnConfiguration.cs
mslukebo: done

src/Microsoft.Performance.SDK.Runtime/DTO/TableConfigurationsSerializer.cs
xj-ms: So sounds like this isn't actually possible with our existing code? Usually it's better to guard potential developer mistakes with tests than adding runtime guard/checks. For this case we can have a parameterized (by the configuration type) test that verify they won't create infinite upgrade cycle (by using timeout)

mslukebo: We have tests that perform a round-trip serialize/deserialize from the pre-v1.0 configuration to the latest, so we are already verifying we don't hit an infinite loop. I am fine removing this check given that we have the tests, but I usually prefer more defensive programming where I assume tests like that **don't** exist. I would prefer a future bug where tables don't get deserialized correctly then a bug where an application hangs indefinitely

mslukebo: Although, if we are fine keeping this, I should probably update this to be smarter than a standard reference equality check. I think the current implementation only works if two returned upgraded versions are the same instance

xj-ms: I would say we remove this. It helps reduce the complexity of the production code given we have the tests

xj-ms: We can leave a comment if you feel it's something that future developers should pay attention to

mslukebo: Changed this to check the version returned by `UpgradeToNext`

xj-ms (approved):

src/Microsoft.Performance.SDK/Processing/TableConfiguration.cs
```diff
@@ -184,11 +184,15 @@ public IEnumerable<ColumnConfiguration> Columns
         ///     Gets the mapping of column roles to column guids, if any.
         ///     If there are no column roles, then this dictionary will
         ///     be empty.
+        ///     Only the base projection of a column can be assigned a role; variants
+        ///     of a column cannot be assigned a role.
```
jmaxson-ms, line 188: Just thinking about docs, this is a good reason to create a new column rather than a variant.

jmaxson-ms (approved):

trgibeau (approved):

PR #381, opened 2024-11-05
Un-consolidate ColumnConfiguration constructors

mslukebo:
Plugins that used the old constructors cannot call the new one with default parameters.

### DIFF
--- a/src/Microsoft.Performance.SDK/Processing/ColumnConfiguration.cs
+++ b/src/Microsoft.Performance.SDK/Processing/ColumnConfiguration.cs
@@ -17,6 +17,43 @@ namespace Microsoft.Performance.SDK.Processing
         /// <param name="metadata">
         ///     The metadata about the column.
         /// </param>
+        /// <exception cref="System.ArgumentNullException">
+        ///     <paramref name="metadata"/> is <c>null</c>.
+        /// </exception>
+        public ColumnConfiguration(ColumnMetadata metadata)
+            : this(metadata, null, null)
+        {
+        }
+
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="ColumnConfiguration"/>
+        ///     class.
+        /// </summary>
+        /// <param name="metadata">
+        ///     The metadata about the column.
+        /// </param>
+        /// <param name="hints">
+        ///     Optional hints about displaying this column in the UI. This parameter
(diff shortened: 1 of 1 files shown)


src/Microsoft.Performance.SDK/Processing/ColumnConfiguration.cs
```diff
@@ -10,6 +10,43 @@ namespace Microsoft.Performance.SDK.Processing
...
+        ///     Initializes a new instance of the <see cref="ColumnConfiguration"/>
+        ///     class.
+        /// </summary>
+        /// <param name="metadata">
+        ///     The metadata about the column.
+        /// </param>
+        /// <exception cref="System.ArgumentNullException">
+        ///     <paramref name="metadata"/> is <c>null</c>.
+        /// </exception>
+        public ColumnConfiguration(ColumnMetadata metadata)
```
xj-ms, line 23: Add comments that certain code expects the exact constructor signature?

mslukebo: I can if you want. We have telescoping constructors in many places in the SDK, so I'm not sure this is particularly special code that warrants a call out though

xj-ms: In my opinion we need some mechanism to prevent accidental changes to all of them. Doesn't necessarily need to be comments but can also be tests or other approaches. For now, we can just make improvements for this one

xj-ms: How well do we distinguish public and internal code? If done well, I think we don't need comments but can always assume that public code cannot be changed

mslukebo: This project does a good job at separating internal and public code. Basically, we (try to) never break back compat for public members inside the `SDK` project. `SDK.Runtime` is fair game to change, though.

These types of changes are difficult to unit test though, because the unit test would be compiling against the same code we are changing. We actually have unit tests that compile against these ctors, and those all worked fine without change because they were compiled against the single ctor with optional parameters.

xj-ms: Then I agree comments are not necessary here

xj-ms (approved):

jmaxson-ms (approved):

PR #382, opened 2024-11-13
Use base column description for base mode's ColumnVariantDescriptor.Description

mslukebo:


xj-ms (dismissed):

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnVariantProperties.cs
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
```
xj-ms, line 3: Should we make the properties required?

xj-ms (dismissed):

xj-ms (approved):

jmaxson-ms (approved):

PR #385, opened 2024-11-22
Add ColumnName property to ColumnVariantProperties

mslukebo:


src/Microsoft.Performance.SDK/Processing/ColumnMetadata.cs
```diff
@@ -160,7 +160,7 @@ public ColumnMetadata(ColumnMetadata other)
         ///     on the data in the first row. Use <see cref="NameProjection"/>
         ///     for dynamic names.
         /// </summary>
-        public string Name { get; }
+        public string Name { get; init; }
```
jmaxson-ms, line 163: >public string Name { get; init; } [](http://example.com/codeflow?start=8&length=33)

do we want a nullorwhitespace check on this init like we have in the constructors?

mslukebo: We can do that, but not without declaring our own backing property :(

src/Microsoft.Performance.SDK/Processing/ColumnMetadata.cs
```diff
@@ -155,12 +156,29 @@ public ColumnMetadata(ColumnMetadata other)
         public Guid Guid { get; }
 
         /// <summary>
-        ///     Gets the name of this column. If this column has a dynamic
+        ///     Gets or initializes the name of this column. If this column has a dynamic
         ///     name, then this property always returns the name as based
         ///     on the data in the first row. Use <see cref="NameProjection"/>
         ///     for dynamic names.
```
xj-ms, line 162: Are there any documents explaining the difference between a column's label and name? Still confused after reading the comments

src/Microsoft.Performance.SDK.Runtime/ColumnBuilding/Builders/EmptyColumnBuilder.cs
```diff
@@ -73,7 +78,13 @@ public override ToggleableColumnBuilder WithHierarchicalToggle<T>(
         return new ToggledColumnBuilder(
             [
                 new ToggledColumnBuilder.AddedToggle(toggleDescriptor,
-                    new HierarchicalDataColumn<T>(baseColumn.Configuration, projection, collectionProvider)),
+                    new HierarchicalDataColumn<T>(
+                        new ColumnConfiguration(this.baseColumn.Configuration)
+                        {
+                            Metadata = new ColumnMetadata(this.baseColumn.Configuration.Metadata) { Name = toggleDescriptor.Properties.ColumnName ?? this.baseColumn.Configuration.Metadata.Name},
```
xj-ms, line 84: Can use the `with` keyword here if `ColumnMetadata` is a record

xj-ms (approved):

src/Microsoft.Performance.SDK.Runtime/ColumnBuilding/Builders/EmptyColumnBuilder.cs
mslukebo: yep, I am aware but didn't want to convert ColumnMetadata

src/Microsoft.Performance.SDK/Processing/ColumnMetadata.cs
mslukebo: Label's are associated with Column Metadata; the label is only associated with the variant.

I think this will be more clear once we have an implementation in an SDK driver that we can use to augment our documentation. The final docs for this feature are still a WIP, so I'm OK with some things still being a bit ill-defined (especially as we are still developing them).

xj-ms: > Label's are associated with Column Metadata

I assume you mean the names are associated with column metadata?

mslukebo: yep, my bad!

trgibeau (approved):

PR #390, opened 2024-12-03
Support exposing timezones relevant to processed datasources

mslukebo:
Currently, the `DataSourceInfo` a data processor exposes requires its `DateTime` be UTC. This constraint may cause a processor to disregard relevant information, such as which time zone the data source was created.

This PR adds a new `DataSourceTimeZone` record that describes time zones of significance to data sources. Instances of these records are now accessible from the `DataSourceInfo`, allowing plugins to expose this previously disregarded information to SDK drivers.

### DIFF
--- a/src/Microsoft.Performance.SDK/Processing/DataSourceTimeZone.cs
+++ b/src/Microsoft.Performance.SDK/Processing/DataSourceTimeZone.cs
@@ -1,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Performance.SDK.Processing;
+
+/// <summary>
+///     Represents a time zone that is of significance to a data source.
+/// </summary>
+/// <param name="Label">
+///     A human-readable label that describes the significance of the time zone.
+/// </param>
+/// <param name="TimeZone">
+///     The time zone.
+/// </param>
+public sealed record DataSourceTimeZone(string Label, TimeZoneInfo TimeZone);
--- a/src/Microsoft.Performance.SDK/Processing/DataSourceInfo.cs
+++ b/src/Microsoft.Performance.SDK/Processing/DataSourceInfo.cs
@@ -136,5 +136,10 @@ namespace Microsoft.Performance.SDK.Processing
                 return this.FirstEventWallClockUtc.AddTicks(this.EndTimestampNanoseconds / 100);
             }
         }
+
+        /// <summary>
+        ///     Gets or initializes the time zones that are of significance to this data source.
+        /// </summary>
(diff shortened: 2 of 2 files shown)


src/Microsoft.Performance.SDK/Processing/DataSourceTimeZone.cs
```diff
@@ -0,0 +1,17 @@
...
+/// <summary>
+///     Represents a time zone that is of significance to a data source.
+/// </summary>
+/// <param name="Label">
+///     A human-readable label that describes the significance of the time zone.
+/// </param>
+/// <param name="TimeZone">
+///     The time zone.
+/// </param>
+public sealed record DataSourceTimeZone(string Label, TimeZoneInfo TimeZone);
```
xj-ms, line 17: Time zones are well-defined. `TimeZoneInfo` already has a `StandardName` property which is also localized by the OS.
For this kind of info I think it's better to let the UI control the format instead of the data sources

mslukebo: This isn't for the time zone; i.e. it's not supposed to be "UTC" or "PST". 

> A human-readable label that describes the significance of the time zone.

This is supposed to describe what the time-zone represents. E.g. "time zone where the trace was recorded"

xj-ms: Should we make it optional? I imagine for most data sources, this is going to be the time zone that the trace is recorded.
Also it should not be involved in the equality computation of the record

mslukebo: > Should we make it optional?

I don't think so. I would rather not have to guess what label we should display if it's null. We also shouldn't make assumptions that the use case we are considering is the most common one.

> Also it should not be involved in the equality computation of the record

Why not?

xj-ms: If 2 data sources happen to have `DataSourceTimeZone` with the same time zone and description, it doesn't make sense to consider them equal (and probably not very useful to do so). Depending on how we want to use it, I think the equality should be either determined by the time zone info or left undefined.

mslukebo: > If 2 data sources happen to have DataSourceTimeZone with the same time zone and description, it doesn't make sense to consider them equal

I'm not sure I agree. If two instances of pure data classes have the same data, I would consider them equal.

If we're going to alter equality, I would be more in favor of just leaving default reference equality. But, as this is a plain data object, I don't see the harm in using a record type with its default equality comparison.

xj-ms (approved):

jmaxson-ms (approved):
:shipit:

PR #391, opened 2024-12-06
Use a dictionary for DataSourceInfo.TimeZones

mslukebo:


src/Microsoft.Performance.SDK/Processing/DataSourceInfo.cs
```diff
@@ -138,8 +138,18 @@ public DateTime EndWallClockUtc
...
+        /// <example>
+        ///     If the data source was a trace file recorded on a machine in the Eastern Time Zone, an entry in this
+        ///     dictionary might be "Trace Local" => EST.
+        /// </example>
+        /// <example>
+        ///     If the data source was a trace recorded on a web server where the host was in the Eastern Time Zone and
+        ///     the client was in the Pacific Time Zone, entries in this dictionary might be "Host Local" => EST and
+        ///     "Client Local" => PST.
+        /// </example>
+        public IReadOnlyDictionary<string, TimeZoneInfo> TimeZones { get; init; } = new Dictionary<string, TimeZoneInfo>();
```
xj-ms, line 153: Should we use the zone as the key? That will allow them to have multiple same time zone with different string values?

mslukebo: > That will allow them to have multiple same time zone with different string values

Isn't that the opposite of what you're suggesting? With the label as the key, we can have 2 different keys with the same time zone.

We want to avoid having one label being associated with the same time zone, which is why the label is the key here.

jmaxson-ms (approved):

xj-ms (approved):

trgibeau (approved):

PR #392, opened 2024-12-10
Move ProcessorEnvironment and factory to SDK.

jmaxson-ms:
This allows a plugin to reference this type without referencing the Engine directly.
 
ProcessorEnvironment also inherits from IProcessorEnvironmentV2, with a default impelmentation..


xj-ms:
Is this a breaking change? If so how large will the impact be?

mslukebo (dismissed):

jmaxson-ms:
> Is this a breaking change? If so how large will the impact be?

This has not been released in a public nuget package yet.

xj-ms (dismissed):

mslukebo (approved):

xj-ms (approved):

PR #393, opened 2025-01-14
Expose new table commands that expose command context and support being unavailable

mslukebo:
This PR adds a new `ITableBuilder.AddTableCommand2` path that allows plugin authors to
1. Retrieve contextual information for how the command is being executed
2. Supply a `CanExecute` predicate that evaluates if the command can be performed

The contextual information applied to both the `CanExecute` and `OnExecute` methods includes
* The selected rows
* The table's current table configuration (if available)
* The `Guid` of the column the command is being executed on (if available)


src/Microsoft.Performance.SDK.Runtime/TableBuilder.cs
```diff
@@ -47,8 +46,10 @@ public TableBuilder()
             this.columns = new List<IDataColumn>();
             this.builtInTableConfigurations = new List<TableConfiguration>();
             this.columnsRO = new ReadOnlyCollection<IDataColumn>(this.columns);
-            this.commands = new List<TableCommand>();
-            this.commandsRO = new ReadOnlyCollection<TableCommand>(this.commands);
+
+            this.commands2 = new List<TableCommand2>();
+            this.Commands = new ReadOnlyCollection<TableCommand2>(this.commands2);
```
xj-ms, line 51: Are there benefits of creating a `ReadOnlyCollection` instead of directly assinging `commands2` to `Commands`?

src/Microsoft.Performance.SDK.Runtime/TableCommand2.cs
```diff
@@ -0,0 +1,24 @@
...
+/// <param name="CanExecute">
+///     A <see cref="Predicate{T}"/> that determines if the command can be executed.
+/// </param>
+/// <param name="OnExecute">
+///     The function to execute when the command is invoked.
+/// </param>
+public sealed record TableCommand2(
+    string CommandName,
+    Predicate<TableCommandContext> CanExecute,
+    Action<TableCommandContext> OnExecute);
```
xj-ms, line 24: Since we have the opportunity, how about also adding support for async operations as well i.e. `Func<TableCommandContext, Task> OnExecute`?

jmaxson-ms (dismissed):
:shipit:

src/Microsoft.Performance.SDK/Processing/ITableBuilder.cs
```diff
@@ -53,6 +53,43 @@ public interface ITableBuilder
         /// </exception>
         ITableBuilder AddTableCommand(string commandName, TableCommandCallback callback);
 
+        /// <summary>
+        ///     Registers a command that users can execute against this table.
+        /// </summary>
+        /// <param name="commandName">
+        ///     The name of the command. This name ultimately shows up in the context
+        ///     menu in the UI when the user right clicks a table. The name of the command
```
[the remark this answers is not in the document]
xj-ms: `commandName` is the name of the context menu item, right? Then it shows when the user right clicks a table but before the command is actually invoked

src/Microsoft.Performance.SDK.Runtime/TableBuilder.cs
mslukebo: Nope, I was just following what was there. Updated

src/Microsoft.Performance.SDK.Runtime/TableCommand2.cs
mslukebo: This would be a bit more of a bigger change that I would like to introduce with this PR. Forcing plugin authors (and SDK drivers) to use async for this would require more refactoring that I would like to force for this feature. I think this is something we might want to do for SDK 2.0 though, when we remove the old table command and refactor more things.

xj-ms: It won't really force them in my opinion. They can still have synchronous commands by just returning Task.CompletedTask.

xj-ms (approved):

src/Microsoft.Performance.SDK.Runtime/TableCommand2.cs
mslukebo: The bigger issue is having SDK drivers (e.g. WPA) using an async path to await the callbacks.

trgibeau (approved):

jmaxson-ms (approved):
:shipit:

Issue #366, opened 2024-07-16
Remove IApplicationEnvironment (v1 and v2) and just use abstract class in the SDK


mslukebo:
Because the SDK targets netstandard2.0, we cannot make use of default interface implementations of methods. This causes headaches when we want to add new functionality to the application environment, because adding a member to the interface is a major breaking change.

If we get rid of the interfaces and just use the abstract class, we can use virtual members to add new features while maintaining backwards compatibility.

PR #400, opened 2025-01-27
Add ability to register persisted plugin options

### LINKED ISSUE #401: Add API for declaring options on an IProcessingSource

mslukebo:
This PR adds initial support for plugin options, where 3 types of options are available to use:
* a boolean option
* a string option
* a list of strings option

Plugin options are declared by plugins via a new property on `IProcessingSource`, `IEnumerable<PluginOption> PluginOptions { get; }`. The abstract class `ProcessingSource` implements this virtually with an empty collection. (Work item #405 was created to track support for defining options in plugins without processing sources).

Plugins can obtain instances of their declared `PluginOption`s via a new method on the application environment, `bool TryGetPluginOption<T>(Guid optionGuid, out T option) where T : PluginOption`. This method was created on a new `IApplicationEnvironmentV3` interface, accessible via an extension method on `IApplicationEnvironment`. 

To prevent plugins from using static `PluginOption` instances throughout a plugin, only cloned `PluginOption` instances are ever returned by `TryGetPluginOption`. This invariant is maintained by a new `PluginOptionsRegistry` class in the runtime, which is passed to the runtime's `ApplicationEnvironment` class and is used to find `PluginOption` instances available during `TryGetPluginOption`. (Note: `IApplicationEnvironment` instances defined outside the SDK may break this invariant; once #366 is finished in SDK 2.0 this will not be a concern). The public API for populating options in the `PluginOptionsRegistry` ensures internal members are used to store the cloned option instances.

To facilitate persisting plugin options, DTO classes were added. Two interfaces for saving and loading these DTO objects, `IPluginOptionsSaver` and `IPluginOptionsLoader`, were created and implemented. The `PluginOptionsRegistry` can be "updated" via these DTO classes, which will alter the state of registered plugin option instances based on the DTO values.

Closes #401, #402, and #403


src/Microsoft.Performance.SDK/Options/PluginOption`1.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+        init
+        {
+            this.defaultValue = value;
+            this.currentValue = value;
+        }
+    }
+
+    /// <summary>
+    ///     Gets the current value of the option.
```
helenkzhang, line 46: nit: or sets

mslukebo: This is a strange case where the getter is public but the setter is internal only. Since plugins can only use the getter, I opted to only document that one.

If you feel strongly about it, I can make an internal method that for setting that is documented accordingly.

helenkzhang: I'm fine with not documenting.

src/Microsoft.Performance.SDK.Runtime/Options/PluginOptionsRegistry.cs
```diff
@@ -0,0 +1,144 @@
...
+            (option, dto) => { option.CurrentValue = dto.Value; });
+    }
+
+    private void ApplyT<T, TDTO>(
+        IReadOnlyCollection<TDTO> dtoOptions,
+        Action<T, TDTO> applySaved)
+        where T : PluginOption
+        where TDTO : PluginOptionDto
+    {
+        foreach (var dto in dtoOptions)
```
helenkzhang, line 132: Enclose foreach with { } ?

helenkzhang:
Can we add new line at EOF for those that don't have one?

src/Microsoft.Performance.SDK.Runtime/Options/PluginOptionsRegistry.cs
mslukebo: Thanks, my IDE auto-formatted this file.

mslukebo:
> Can we add new line at EOF for those that don't have one?

There are a lot of other files in this repo that also are missing EOF. I would prefer to not bother going through all of individual files in this PR. Instead we should do a separate PR adding editor config that checks this stuff for us.

helenkzhang (dismissed):

src/Microsoft.Performance.SDK.Runtime/Microsoft.Performance.SDK.Runtime.csproj
```diff
@@ -40,4 +40,8 @@
     <ProjectReference Include="..\Microsoft.Performance.SDK\Microsoft.Performance.SDK.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Text.Json" Version="8.0.5" />
```
jmaxson-ms, line 44: >Version="8.0.5" [](http://example.com/codeflow?start=49&length=15)

should we put this in directory.build.targets?

jmaxson-ms: actually, do we need it?

src/Microsoft.Performance.SDK/Processing/IProcessingSource.cs
```diff
@@ -43,6 +44,11 @@ public interface IProcessingSource
         /// </summary>
         IEnumerable<Option> CommandLineOptions { get; }
 
+        /// <summary>
+        ///     Gets the plugin options that are supported by this <see cref="IProcessingSource"/>.
+        /// </summary>
+        IEnumerable<PluginOption> PluginOptions { get; }
+
```
jmaxson-ms, line 51: breaking change? do we care?

jmaxson-ms:
>             this.InitializeAllTables(this.ApplicationEnvironment.Serializer);

are these changes intentional? they don't seem related to options.

---
Refers to: src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs:373 in e1c8f5d. [](commit_id = e1c8f5d34eb308fb2d5136c3026d2c393a6be415, deletion_comment = True)

src/Microsoft.Performance.SDK/Options/PluginOption.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    /// <summary>
+    ///     Raised when this option's value changes.
+    /// </summary>
+    public event EventHandler OptionChanged;
+
+    /// <summary>
+    ///     Initializes a new instance of the <see cref="PluginOption"/> class.
+    /// </summary>
+    internal PluginOption()
```
jmaxson-ms, line 22: >internal [](http://example.com/codeflow?start=4&length=8)

private protected?

src/Microsoft.Performance.SDK/Options/PluginOption.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// <summary>
+    ///     Gets or initializes the human-readable description of the option. This value is required to be initialized.
+    /// </summary>
+    public required string Description { get; init; }
+
+    /// <summary>
+    ///     Gets a value indicating whether this option is using its default value. This value will be <c>true</c>
+    ///     if the option's value is manually set outside of <see cref="ApplyDefault"/>, regardless of which
+    ///     value it was set to. This means that this value will be <c>true</c> even if the option's value is
+    ///     manually set to its default value.
```
jmaxson-ms, line 51: Am I just reading this wrong? Should this be <c>false</c>

src/Microsoft.Performance.SDK/Processing/IProcessingSource.cs
mslukebo: Breaking change, but we decided when we did previous features that we're OK breaking people who directly implement IProcessingSource. For example, we broke this interface when we added the `IDataSourceGroup` functionality.

src/Microsoft.Performance.SDK/Options/PluginOption.cs
mslukebo: Thanks, that's a typo. I trusted copilot's auto complete!

src/Microsoft.Performance.SDK.Runtime/Options/PluginOptionsRegistry.cs
```diff
@@ -0,0 +1,144 @@
...
+{
+    private readonly object mutex = new();
+    private readonly Dictionary<Guid, PluginOption> optionByGuid = new();
+
+    /// <summary>
+    ///     Represents a class that can provide <see cref="PluginOption"/> instances to register to this class. Do not
+    ///     implement this interface directly. Instead, use public methods on <see cref="PluginOptionsSystem"/> to register
+    ///     options from supported sources.
+    /// </summary>
+    internal interface Provider
```
jmaxson-ms, line 25: >Provider [](http://example.com/codeflow?start=23&length=8)

an interface that doesn't start with 'I'. Is that something we really want?

mslukebo:
> > ```
> >         this.InitializeAllTables(this.ApplicationEnvironment.Serializer);
> > ```
> 
> are these changes intentional? they don't seem related to options.
> 
> Refers to: src/Microsoft.Performance.SDK/Processing/ProcessingSource.cs:373 in [e1c8f5d](https://github.com/microsoft/microsoft-performance-toolkit-sdk/commit/e1c8f5d34eb308fb2d5136c3026d2c393a6be415). [](commit_id = [e1c8f5d](https://github.com/microsoft/microsoft-performance-toolkit-sdk/commit/e1c8f5d34eb308fb2d5136c3026d2c393a6be415), deletion_comment = True)

This change was extracted into a separate PR (#399). Not sure why you're seeing it here since GitHub's UI doesn't show the change for me:

image

src/Microsoft.Performance.SDK.Runtime/Options/PluginOptionsRegistry.cs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    private void ApplyT<T, TDTO>(
+        IReadOnlyCollection<TDTO> dtoOptions,
+        Action<T, TDTO> applySaved)
+        where T : PluginOption
+        where TDTO : PluginOptionDto
+    {
+        foreach (var dto in dtoOptions)
+            lock (mutex)
```
jmaxson-ms: >lock (mutex) [](http://example.com/codeflow?start=12&length=12)

we probably don't need to worry about a lock convoy, but it still might be better to take this once outside of the for loop

src/Microsoft.Performance.SDK.Runtime/Options/PluginOptionsRegistry.cs
mslukebo: I'll change it, but I left it off since this is nested in a class. `TestPluginOptionsRegistryProvider : PluginOptionsRegistry.Provider` looked better to me than `TestPluginOptionsRegistryProvider : PluginOptionsRegistry.IProvider` 🤷

jmaxson-ms: Thanks, I think the consistency will make it easier to read.

src/Microsoft.Performance.SDK.Runtime/Options/Serialization/Loading/StreamPluginOptionsLoader.cs
```diff
@@ -0,0 +1,65 @@
...
+        };
+
+        var stream = GetStream();
+        try
+        {
+            return await JsonSerializer.DeserializeAsync<PluginOptionsDto>(stream, jsonSerializerOptions);
+        }
+        catch
+        {
+            return null;
```
jmaxson-ms, line 47: >return null; [](http://example.com/codeflow?start=12&length=12)

do we want any sort of logging for this failure case?

mslukebo: Yes - my next PR that I have stacked on this branch will add logging.

src/Microsoft.Performance.SDK.Runtime/Options/Serialization/Saving/StreamPluginOptionsSaver.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        var stream = GetStream();
+        try
+        {
+            await JsonSerializer.SerializeAsync(stream, optionsDto, jsonSerializerOptions);
+            return true;
+        }
+        catch
+        {
+            return false;
```
jmaxson-ms, line 48: >return false; [](http://example.com/codeflow?start=12&length=13)

do we want any logging for this?

mslukebo: Yep, will be added in next PR

jmaxson-ms (approved):

helenkzhang (approved):

src/Microsoft.Performance.SDK.Runtime/Options/Serialization/Loading/StreamPluginOptionsLoader.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Microsoft.Performance.SDK.Runtime.Options.Serialization.Loading;
+
+/// <summary>
+///     Base class for <see cref="IPluginOptionsLoader"/> classes which load from a stream.
+/// </summary>
+public abstract class StreamPluginOptionsLoader
+    : IPluginOptionsLoader
+{
+    private readonly bool closeStreamOnWrite;
```
trgibeau, line 17: nit: closeStreamOnRead

src/Microsoft.Performance.SDK.Runtime/Options/Serialization/Saving/StreamPluginOptionsSaver.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    /// <inheritdoc />
+    public async Task<bool> TrySaveAsync(PluginOptionsDto optionsDto)
+    {
+        var jsonSerializerOptions = new JsonSerializerOptions
+        {
+            WriteIndented = true,
+            PropertyNameCaseInsensitive = true,
+        };
+
```
trgibeau, line 39: Should we have these options in a common place to be used by both the loader and saver?

src/Microsoft.Performance.SDK/Options/FieldOption.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Microsoft.Performance.SDK.Options;
+
+/// <summary>
+///     A <see cref="PluginOption{T}"/> for a single string field.
+/// </summary>
+public sealed class FieldOption
+    : PluginOption<string>
+{
+    public override PluginOption CloneT()
```
trgibeau, line 12: inherit doc?

src/Microsoft.Performance.SDK/Options/PluginOption`1.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// </summary>
+    private protected PluginOption()
+    {
+        this.IsUsingDefault = true;
+    }
+
+    /// <summary>
+    ///     Gets or initializes the default value of the option. This value is required to be initialized.
+    /// </summary>
+    public required T DefaultValue
```
xj-ms, line 32: >DefaultValue [](http://example.com/codeflow?start=22&length=12)

I think default values should not be defined by an option type. It will make reuse easier if users can define different default values for the same type of option under different circumstances

src/Microsoft.Performance.SDK/Options/PluginOption.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// <summary>
+    ///     Gets or initializes the <see cref="Guid"/> of the option. This value is required to be
+    ///     initialized.
+    /// </summary>
+    public required Guid Guid { get; init; }
+
+    /// <summary>
+    ///     Gets or initializes the human-readable category of the option. This value is required to be initialized.
+    /// </summary>
+    public required string Category { get; init; }
```
xj-ms, line 35: >required [](http://example.com/codeflow?start=11&length=8)

I imagine most plugins won't have a lot of options that may benefit from having multiple categories. In such case we could define a general category for the developers instead of requiring them to specify one

src/Microsoft.Performance.SDK.Runtime/Options/PluginOptionsRegistry.cs
```diff
@@ -0,0 +1,150 @@
...
+        ///     Gets the <see cref="PluginOption"/> instances to register.
+        /// </summary>
+        /// <returns>
+        ///     The <see cref="PluginOption"/> instances to register.
+        /// </returns>
+        IEnumerable<PluginOption> GetOptions();
+    }
+
+    /// <summary>
+    ///     Gets the <see cref="PluginOption"/> instances that have been registered.
```
xj-ms, line 39: >Gets the <see cref="PluginOption"/> instances that have been registered. [](http://example.com/codeflow?start=12&length=72)

Since this is public, maybe we should specify the fact that the returned collection is a snapshot that won't update if the registry is updated

src/Microsoft.Performance.SDK/Options/PluginOption.cs
xj-ms: I guess it depends on whether we group the options by plugins or not

src/Microsoft.Performance.SDK.Runtime/Options/PluginOptionsRegistry.cs
mslukebo: Will update in next PR.

src/Microsoft.Performance.SDK/Options/PluginOption.cs
mslukebo: Resolved offline.

src/Microsoft.Performance.SDK/Options/PluginOption`1.cs
mslukebo: Resolved offline.

src/Microsoft.Performance.SDK.Runtime/Options/PluginOptionsSystem.cs
```diff
@@ -0,0 +1,147 @@
...
+using Microsoft.Performance.SDK.Runtime.Options.Serialization;
+using Microsoft.Performance.SDK.Runtime.Options.Serialization.Loading;
+using Microsoft.Performance.SDK.Runtime.Options.Serialization.Saving;
+
+namespace Microsoft.Performance.SDK.Runtime.Options;
+
+/// <summary>
+///     Represents a system for managing plugin options.
+/// </summary>
+public sealed class PluginOptionsSystem
```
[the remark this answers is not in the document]
mslukebo: We're semi-consistent in this solution with using "System" for classes that are mostly just a composition of different interfaces. E.g. we have the `PluginsSystem` in the plugins runtime. The "Manager" and "Service" classes in this solution are more focused in their scope.

PR #410, opened 2025-02-06
User/lukebo/plugin options changes

mslukebo:
* Makes `PluginOptionsRegistry.UpdateFromDto` a public method so that runtime users can manually update options without forcing re-loading from the loader.
* Ensures that saving the current registry does not erase previously saved values.
    * Adds tests for this case
* `FilePluginOptionsLoader` now opens the file stream with `OpenOrCreate`


jmaxson-ms (approved):
:shipit:

PR #411, opened 2025-02-07
Plugin options system improvements

mslukebo:
* Prevents a NRE when calling `PluginOptionsSystem.TrySaveCurrentRegistry` if the loader fails to load options. Adds a test for this scenario.
* Updates `StreamPluginOptionsLoader` to return an empty DTO if the stream being loaded from has no data. Adds a test for this scenario.
* Makes the filepaths on the file loader/saver public.

### DIFF
--- a/src/Microsoft.Performance.SDK.Runtime/Options/Serialization/Loading/FilePluginOptionsLoader.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/Options/Serialization/Loading/FilePluginOptionsLoader.cs
@@ -12,10 +12,8 @@ namespace Microsoft.Performance.SDK.Runtime.Options.Serialization.Loading;
 ///     A <see cref="IPluginOptionsLoader"/> that can load a <see cref="PluginOptionsDto"/> instance from a file.
 /// </summary>
 public sealed class FilePluginOptionsLoader
-    : StreamPluginOptionsLoader
+    : StreamPluginOptionsLoader<FileStream>
 {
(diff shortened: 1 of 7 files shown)


jmaxson-ms (approved):
:shipit:

src/Microsoft.Performance.SDK.Runtime/Options/Serialization/Loading/FilePluginOptionsLoader.cs
```diff
@@ -37,17 +35,28 @@ public FilePluginOptionsLoader(string filePath, ILogger logger)
...
+    {
+        return File.Open(this.FilePath, FileMode.OpenOrCreate, FileAccess.Read, FileShare.Read);
     }
 
     /// <inheritdoc />
-    protected override Stream GetStream()
+    protected override bool HasContent(FileStream stream)
     {
-        return File.Open(this.filePath, FileMode.OpenOrCreate, FileAccess.Read, FileShare.Read);
+        return stream.Length > 0;
```
xj-ms, line 60: The 2 overrides are the same and can be done in the base class?

mslukebo: They are, but not every `Stream` supports getting the length. If the returned stream doesn't support it, getting `Length` would throw an exception. I would rather make implementors figure out how to determine if the stream is empty for those types of streams.

PR #417, opened 2025-03-18
Use SingleOrDefault in GetPrebuiltConfigurationFromEmbeddedResource

xj-ms:
This starts to throw with the newly added AI tables because `GetManifestResourceNames` is empty. Not sure if this is a sign of a bigger issue. If this is expected, this change will stop it from throwing exceptions unnecessarily


mslukebo (dismissed):

xj-ms:
@jmaxson-ms Updated. Also updated another place where `e.Message` was used

mslukebo (approved):

PR #418, opened 2025-03-19
Validate plugin identifiers

mslukebo:


src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/PluginIdentity.cs
```diff
@@ -88,5 +90,34 @@ public override string ToString()
         {
             return $"{this.Id}-{this.Version}";
         }
+
+        public bool HasValidId(out string errorMessage)
```
xj-ms, line 94: Maybe add a comment to point out that this is to ensure it can be used as a file path

xj-ms (approved):

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/PluginIdentity.cs
mslukebo: Updated

jmaxson-ms:
This could potentially break existing plugins? I assume this isn't going to cause too many headaches?

mslukebo:
> This could potentially break existing plugins? I assume this isn't going to cause too many headaches?

In theory yes. In practice, any plugins that violated these constraints most likely were throwing exceptions during installation anyway

jmaxson-ms (approved):

PR #419, opened 2025-03-27
Separate plugin option surface API into separate classes

mslukebo:
While writing the documentation for plugin options, it was clear that the current API would be confusing for plugin authors. Specifically, the current design has `PluginOption` classes defined in the SDK namespace that are used to
1. Define options on the `ProcessingSource`
2. Fetch option values from within the plugin
3. Set values from within the SDK or SDK driver (or potentially via loaded plugins!)

The `PluginOption` classes are doing too much work. It is also confusing that a plugin author must be careful to only use `PluginOption` instances returned via the application environment for fetching values. It is possible for an author to accidentally re-use the same instance returned from a `ProcessingSource` and be confused as to why it's not receiving values.

This PR refactors the API with the following:
1. The SDK namespace now exposes two classes:
    1. `PluginOptionDefinition` classes which are used to **define** options on a `ProcessingSource`. These classes are plain data objects and contain the name, description, default value, etc.. Crucially, these classes do not offer a way to retrieve option values.
    2. `PluginOptionValue` classes which are used to **access option values**. Plugins can obtain instances of these classes are obtained exclusively via the application environment. These classes are publicly read-only, meaning other plugins cannot update their values.
 
Finally, the old `PluginOption` classes were moved into the SDK Runtime namespace. The classes are still exposed via the `PluginOptionsRegistry` so that SDK drivers can mutate option values. These classes now contain
1. Methods for setting the option value and re-applying their defaults
2. References to the `PluginOptionDefinition` and `PluginOptionValue` that make up the runtime option.
3. A boolean indicator of if the option is using the default value. This bool is not present on the `PluginOptionValue` since plugins should not care about whether a value is the default value.

These changes will be incompatible with the currently released SDK 1.4 preview, causing plugins that use the current implementation to no longer load.


src/Microsoft.Performance.SDK.Runtime/Options/PluginOptionVisitorExecutor.cs
```diff
@@ -64,4 +64,4 @@ private void Visit(PluginOption option)
                 break;
         }
     }
```
xj-ms, line 66: Looking at this again, does the executor defeat the purpose of using the visitor pattern? There won't be a build error if we (or SDK developers if they wish) add a new option type without editing this class

xj-ms:
> ...since plugins should not care about whether a value is the default value.

My only concern is that since they already have access to the default value, they might compare it with the current value to determine whether it's using the default value if for some weird reason they want to know that. However, this is the wrong approach according to our design. They won't know they are wrong, and we won't know they actually are doing it

xj-ms (dismissed):

src/Microsoft.Performance.SDK.Runtime/Options/PluginOptionVisitorExecutor.cs
mslukebo: This is by design. We want the ability to add new option types without requiring a major version bump. If we put the Visitor inside the SDK namespace, we wouldn't be able to add new methods for new types without a major version bump. By putting it in the runtime, we're allowed to break the interface as much as we want.

Because the visitor is in the runtime namepsace, we can't reference the type in the SDK namespace. This means the `PluginOptionDefinition` classes need an executor in the runtime. I'd prefer to follow the same pattern for these classes defined in the runtime also, just to be consistent.

Do you know a better solution?

mslukebo:
> > ...since plugins should not care about whether a value is the default value.
> 
> My only concern is that since they already have access to the default value, they might compare it with the current value to determine whether it's using the default value if for some weird reason they want to know that. However, this is the wrong approach according to our design. They won't know they are wrong, and we won't know they actually are doing it

This is a valid concern. IMO, plugins shouldn't be trying to determine/use this information. I will add a remark to the `CurrentValue` property to call out that you shouldn't do this.

src/Microsoft.Performance.SDK.Runtime/Options/PluginOptionVisitorExecutor.cs
mslukebo: Also, in regards to "or plugin developers if they wish":

We explicitly do not allow plugin developers to define new option types. The constructor for the types in the SDK namespace is `private protected`, meaning it can only be called by types in our DLL (at least for normal compilation). So, this isn't a supported scenario.

mslukebo: @xj-ms I added some unit tests that use reflection to verify each concrete type is handled by our visitors' executors

xj-ms (approved):

PR #423, opened 2025-04-02
Use json converters collection instead of property converters

helenkzhang:


src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Program.cs
```diff
@@ -51,6 +51,10 @@ private static PluginsCli CreatPluginsCli()
                 {
                     WriteIndented = true,
                     PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+                    Converters =
+                    {
+                        new PluginVersionJsonConverter(),
```
mslukebo, line 56: do we need SemanticVersionJsonConverter here too?

Also, should we just have a common place to initialize all these serializers for both areas?

helenkzhang: No, PluginManifest doesn't have a property of SemanticVersion. 

We already have SerializationUtils which handles the initializations of these serializers. I don't want to also move the default options for PluginManifest to Plugins.Core because PluginManifest is only used in the Cli. Or do you have any suggestions?

mslukebo: I just don't want to accidentally add a SemanticVersion property in PluginManifest down the line and accidentally break things again because we assume the converters are the same.

I think as long as we have round-trip tests for both PluginManifest and PluginMetadata/PluginIdentity de/serialization it should be OK. Those tests would catch those scenarios.

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core.Tests/PluginMetadataSerializationTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        stream.Position = 0;
+
+        PluginMetadata deserializedMetadata = metadataSerializer.Deserialize(stream);
+
+        Assert.IsNotNull(deserializedMetadata);
+        Assert.IsTrue(metadata.Equals(deserializedMetadata));
+    }
+
+    [TestMethod]
+    public void PluginContentsMetadata_Serialization_RoundTrip()
```
mslukebo, line 39: These two methods are very similar. You can create a helper method that uses a generic parameter to actually run the test.

```cs
private void RunSerializationTest<T>()
{
        ISerializer<T> serializer = SerializationUtils.GetJsonSerializerWithDefaultOptions<PluginMetadata>();

        T original = this.fixture!.Create<T>();

        using var stream = new MemoryStream();
        serializer.Serialize(stream, original);

        stream.Position = 0;

        T deserialized = serializer.Deserialize(stream);

        Assert.IsNotNull(deserialized);
        Assert.IsTrue(original.Equals(deserialized));
}
```

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/PluginIdentity.cs
```diff
@@ -84,8 +84,8 @@ public static bool Equals(PluginIdentity a, PluginIdentity b)
         public override int GetHashCode()
         {
             return HashCodeUtils.CombineHashCodeValues(
-               this.Id.GetHashCode(),
-               this.Version.GetHashCode());
+               this.Id?.GetHashCode() ?? 0,
```
mslukebo, line 87: Are these actually allowed to be null? Or is this just because the auto fixture isn't actually passing values?

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli.Tests/PluginManifestSerializationTests.cs
```diff
@@ -0,0 +1,39 @@
...
+    [TestInitialize]
+    public void Setup()
+    {
+        this.fixture = new Fixture();
+    }
+
+    [TestMethod]
+    public void PluginManifest_Serialization_RoundTrip()
+    {
+        ISerializer<PluginManifest> manifestSerializer = SerializationUtils.GetJsonSerializer<PluginManifest>(Program.PluginManifestSerializerDefaultOptions);
```
mslukebo, line 25: Same comment here about duplicating test code. Is there a way we can re-use the logic between the two files?

src/PluginsSystem/Tools/Microsoft.Performance.Toolkit.Plugins.Cli/Manifest/PluginIdentityManifest.cs
```diff
@@ -38,5 +38,54 @@ public PluginIdentityManifest(
...
+
+            return string.Equals(a.Id, b.Id, StringComparison.Ordinal) &&
+                PluginVersion.Equals(a.Version, b.Version);
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            return HashCodeUtils.CombineHashCodeValues(
+               this.Id?.GetHashCode() ?? 0,
```
mslukebo, line 87: same comment here

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core/PluginIdentity.cs
helenkzhang: Removed as are not allowed to be null. I added guards in the other file.

src/PluginsSystem/Microsoft.Performance.Toolkit.Plugins.Core.Tests/PluginMetadataSerializationTests.cs
helenkzhang: Added a helper method

mslukebo (approved):

xj-ms (approved):

trgibeau (approved):

PR #427, opened 2025-07-03
defer array allocation until the first use in CachedOnFirstUseColumnGenerator

xj-ms:
The arrays are allocated as soon as a trace is loaded in WPA, but not actually used until a specific column is shown in the analysis view. If there are a lot of rows, several hundred megabytes can be saved by not allocating these arrays

### DIFF
--- a/src/Microsoft.Performance.SDK/Processing/Projection/Projection.cs
+++ b/src/Microsoft.Performance.SDK/Processing/Projection/Projection.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


xj-ms:
@microsoft-github-policy-service agree company="Microsoft"

src/Microsoft.Performance.SDK/Processing/Projection/Projection.cs
```diff
@@ -1960,39 +1960,30 @@ private struct CachedOnFirstUseColumnGenerator<T, TGenerator>
...
+            private readonly Lazy<Cache> lazyCache;
             private TGenerator generator;
 
             public CachedOnFirstUseColumnGenerator(int rowCount, TGenerator generator)
             {
-                this.cache = new T[rowCount];
-                this.isCached = new byte[rowCount];
+                this.rowCount = rowCount;
                 this.generator = generator;
+                this.lazyCache = new Lazy<Cache>(() => new Cache(new T[rowCount], new byte[rowCount]), isThreadSafe: false);
```
mslukebo, line 1971: From the docs:
> If you intend to access the property from just one thread, pass in false to obtain a modest performance benefit

I think we are definitely capable of being accessed from multiple threads. Instead maybe we should use `LazyThreadSafetyMode.ExecutionAndPublication`?

src/Microsoft.Performance.SDK/Processing/Projection/Projection.cs
```diff
@@ -2038,6 +2029,26 @@ public bool NotifyVisibleDomainChanged(IVisibleDomainRegion newVisibleDomain)
             }
 
             public bool DependsOnVisibleDomain => this.generator.DependsOnVisibleDomain();
+
+            private readonly record struct Cache(T[] Values, byte[] IsCached)
```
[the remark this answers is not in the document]
xj-ms: Yes

src/Microsoft.Performance.SDK/Processing/Projection/Projection.cs
xj-ms: Updated. `ExecutionAndPublication` is the default option

mslukebo (approved):

jmaxson-ms (approved):

PR #431, opened 2025-11-27
Allow multiple FileDataSourceAttribute to be used

xj-ms:
Simply removing this will allow plugins to support multiple different file extensions. WPA already has the support for it as WPA always processes `FileDataSourceAttribute` in an `IEnumerable`. Even in this repo we have tests for multiple extensions (`TryGetFileExtensions_FileDataSources_ReturnsExtensions`)

### DIFF
--- a/src/Microsoft.Performance.SDK/Processing/FileDataSourceAttribute.cs
+++ b/src/Microsoft.Performance.SDK/Processing/FileDataSourceAttribute.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Performance.SDK.Processing
     ///     files of a specific type. This is used to instruct callers to route files
     ///     with a given extension to the decorated <see cref="IProcessingSource"/> for processing.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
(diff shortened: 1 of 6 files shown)


malavshah11 (approved):

src/Microsoft.Performance.SDK/Processing/FileDataSourceAttribute.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.Performance.SDK.Processing
     ///     files of a specific type. This is used to instruct callers to route files
     ///     with a given extension to the decorated <see cref="IProcessingSource"/> for processing.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
```
helenkzhang, line 15: Would this introduce any potential issue? Wonder why we did not allow this before.

xj-ms: Did some tests with 2 plugins supporting the same set of file extensions and didn't find any issues
I don't know why it's like this. This is disabled since this file was committed in 2020

helenkzhang (approved):

PR #432, opened 2025-12-10
fix string comparison structs crash when compared with default values

xj-ms:
The string comparison structs are meant to be used as column data types. When their default values are used, the internal string will be null, which triggers a null reference exception in their comparison logic or ToString.

This PR fixes this issue, adds null annotations and unit tests

### DIFF
--- a/src/Microsoft.Performance.SDK.Tests/AssemblyInfo.cs
+++ b/src/Microsoft.Performance.SDK.Tests/AssemblyInfo.cs
@@ -1,0 +1,6 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Performance.Testing;
+
+[assembly: UnitTest]
--- a/src/Microsoft.Performance.SDK.Tests/StringWithCaseInsensitiveComparisonTests.cs
+++ b/src/Microsoft.Performance.SDK.Tests/StringWithCaseInsensitiveComparisonTests.cs
@@ -1,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#nullable enable
+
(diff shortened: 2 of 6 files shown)


src/Microsoft.Performance.SDK.Tests/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Performance.Testing;
+
+[assembly: UnitTest]
```
helenkzhang: nit: add new line at eof

src/Microsoft.Performance.SDK.Tests/StringWithCaseInsensitiveComparisonTests.cs
```diff
@@ -0,0 +1,35 @@
...
+    public void DefaultValue_CompareToBehavesConsistently()
+    {
+        var defaultValue = default(StringWithCaseInsensitiveComparison);
+        var other = new StringWithCaseInsensitiveComparison("apple");
+
+        Assert.AreEqual(0, defaultValue.CompareTo(default(StringWithCaseInsensitiveComparison)));
+        Assert.IsTrue(defaultValue < other);
+        Assert.IsTrue(other > defaultValue);
+    }
+}
```
helenkzhang: Same here

helenkzhang (dismissed):
Approved with suggestions

src/Microsoft.Performance.SDK.Tests/StringWithCaseInsensitiveComparisonTests.cs
xj-ms: Fixed

jmaxson-ms (approved):

helenkzhang (approved):

PR #437, opened 2026-07-17
Introduce ITableBuilder.AddTableCommand3

jmaxson-ms:
Adds ITableBuilder.AddTableCommand3 to remove limitations for:
1. HierarchicalColumns (sub-rows)
2. Allow for return values.

### DIFF
--- a/src/Microsoft.Performance.SDK/Processing/SelectedTableRow.cs
+++ b/src/Microsoft.Performance.SDK/Processing/SelectedTableRow.cs
@@ -1,0 +1,46 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Performance.SDK.Processing;
+
+/// <summary>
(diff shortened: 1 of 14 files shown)


src/Microsoft.Performance.SDK/Processing/SelectedTableRow.cs
```diff
@@ -0,0 +1,46 @@
...
+///     row itself is selected. Never <c>null</c>; a <c>null</c> value passed to the
+///     constructor is normalized to an empty list.
+/// </param>
+public record SelectedTableRow(int RowIndex, IReadOnlyList<int> SubRowIndices)
+{
+    /// <summary>
+    ///     The zero-based indices of the selected sub-rows within the expanded
+    ///     hierarchical column of <see cref="RowIndex"/>. Never <c>null</c>.
+    /// </summary>
+    public IReadOnlyList<int> SubRowIndices { get; init; } = SubRowIndices ?? Array.Empty<int>();
```
xj-ms, line 33: I don't quite understand how a table is going to make use of this. Does it not need to know which column it is?

PR #439, opened 2026-07-28
Add support for table commands with return data - and DownloadSourceCodeCommand.

jmaxson-ms:
This branch introduces a new generation of table commands (TableCommand3) that supports async execution, strongly-typed contexts and results, and column-scoped commands. It also adds the first concrete TableCommand3: DownloadSourceCode. Existing ITableCommand / ITableCommand2 continue to work via adapter types now hosted in the SDK.

### DIFF
--- a/src/Microsoft.Performance.SDK.Runtime/TableBuilder.cs
+++ b/src/Microsoft.Performance.SDK.Runtime/TableBuilder.cs
@@ -1,15 +1,17 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using Microsoft.Performance.SDK.Processing;
+using Microsoft.Performance.SDK.Processing.ColumnBuilding;
+using Microsoft.Performance.SDK.Processing.TableCommands;
+using Microsoft.Performance.SDK.Runtime.ColumnBuilding.Builders;
+using Microsoft.Performance.SDK.Runtime.ColumnBuilding.Processors;
+using Microsoft.Performance.SDK.Runtime.ColumnVariants.Registrar;
+using Microsoft.Performance.SDK.Runtime.TableCommands;
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
-using Microsoft.Performance.SDK.Processing;
-using Microsoft.Performance.SDK.Processing.ColumnBuilding;
-using Microsoft.Performance.SDK.Runtime.ColumnBuilding.Builders;
-using Microsoft.Performance.SDK.Runtime.ColumnBuilding.Processors;
-using Microsoft.Performance.SDK.Runtime.ColumnVariants.Registrar;
 
 namespace Microsoft.Performance.SDK.Runtime
 {
@@ -31,7 +33,7 @@ namespace Microsoft.Performance.SDK.Runtime
         private readonly List<IDataColumn> columns;
         private readonly ReadOnlyCollection<IDataColumn> columnsRO;
         private readonly List<TableConfiguration> builtInTableConfigurations;
-        private readonly List<TableCommand2> commands2;
+        private readonly List<TableCommand3> commands;
         private readonly ColumnVariantsRegistrar variantsRegistrar;
 
         // Maps a row to a collection of row detail entry
@@ -47,8 +49,8 @@ namespace Microsoft.Performance.SDK.Runtime
             this.builtInTableConfigurations = new List<TableConfiguration>();
             this.columnsRO = new ReadOnlyCollection<IDataColumn>(this.columns);
 
-            this.commands2 = new List<TableCommand2>();
-            this.Commands = this.commands2;
+            this.commands = new List<TableCommand3>();
+            this.Commands = this.commands.AsReadOnly();
 
             this.variantsRegistrar = new ColumnVariantsRegistrar();
 
@@ -68,7 +70,14 @@ namespace Microsoft.Performance.SDK.Runtime
         /// <inheritdoc />
         public int RowCount { get; private set; }
 
-        public IReadOnlyList<TableCommand2> Commands { get; }
+        /// <summary>
+        ///     Gets the collection of commands registered against this table.
+        ///     Commands are stored as instances of the abstract
+        ///     <see cref="TableCommand3"/> base type; hosts should inspect the
+        ///     runtime type of each command to determine whether they recognize
+        ///     it. Unknown command types may be safely ignored.
+        /// </summary>
+        public IReadOnlyList<TableCommand3> Commands { get; }
 
         /// <summary>
         ///     Gets the <see cref="IColumnVariantsRegistrar"/> that can be used to find
@@ -94,14 +103,13 @@ namespace Microsoft.Performance.SDK.Runtime
         }
 
         /// <inheritdoc />
-        public ITableBuilder AddTableCommand(string name, TableCommandCallback callback)
+        public ITableBuilder AddTableCommand(string commandName, TableCommandCallback callback)
         {
+            Guard.NotNullOrWhiteSpace(commandName, nameof(commandName));
             Guard.NotNull(callback, nameof(callback));
 
-           return AddTableCommand2(
-               name,
-               (_) => true,
-               (context) => callback(context.SelectedRows));
+            AddCommandCore(new TableCommandCallbackAdapter(commandName, callback));
+            return this;
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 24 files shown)


src/Microsoft.Performance.Toolkit.Engine/RuntimeExecutionResults.cs
```diff
@@ -533,6 +534,9 @@ private sealed class TableBuilder
             private readonly ReadOnlyDictionary<string, TableCommandCallback> tableCommandsRO;
 
             private readonly List<TableCommand2> tableCommands2;
+            private readonly ReadOnlyCollection<TableCommand2> tableCommands2RO;
```
xj-ms, line 537: nit: unnecessary private field

src/Microsoft.Performance.SDK/Processing/TableCommands/TableCommand3.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    /// <summary>
+    ///     The non-generic base for all table commands surfaced through
+    ///     <see cref="ITableBuilder"/>. This base carries the metadata needed to
+    ///     store heterogeneous commands in a single collection. Hosts should
+    ///     inspect the runtime type of each command to determine whether they
+    ///     recognize (and can execute) it. Unknown command types may be safely
+    ///     ignored.
+    /// </summary>
+    public abstract class TableCommand3
```
xj-ms, line 16: I think we should just change this to an interface. Easier to mock when testing

src/Microsoft.Performance.SDK/Processing/TableCommands/ITableCommand3Result.cs
```diff
@@ -0,0 +1,17 @@
...
+    ///     Marker interface implemented by types that represent the result
+    ///     produced by executing a <see cref="TableCommand3{TContext, TResult}"/>.
+    /// </summary>
+    /// <remarks>
+    ///     Implementations convey any output data or status information that
+    ///     the host needs after the command runs. Use
+    ///     <see cref="VoidTableCommandResult"/> for commands that do not
+    ///     produce a meaningful result.
+    /// </remarks>
+    public interface ITableCommand3Result;
```
xj-ms, line 16: Is this necessary? If we don't have any constraint on it, why not just allow all types?

src/Microsoft.Performance.SDK/Processing/TableCommands/TableCommand3`2.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <param name="commandName">
+        ///     The human-readable name of the command.
+        /// </param>
+        protected TableCommand3(string commandName)
+            : base(commandName)
+        {
+        }
+
+        /// <inheritdoc />
+        public override Type ContextType => typeof(TContext);
```
xj-ms, line 44: How is this used?

src/Microsoft.Performance.SDK/Processing/TableCommands/TableCommand3`2.cs
```diff
@@ -0,0 +1,77 @@
...
+    ///     <see cref="ExecuteAsync"/>. Implementations define the shape of this
+    ///     context, allowing commands to receive command-specific state from the
+    ///     host.
+    /// </typeparam>
+    /// <typeparam name="TResult">
+    ///     The type of result produced by <see cref="ExecuteAsync"/>. Use
+    ///     <see cref="VoidTableCommandResult"/> when the command does not
+    ///     produce a value.
+    /// </typeparam>
+    public abstract class TableCommand3<TContext, TResult>
```
xj-ms, line 26: I think this should be an interface too

src/Microsoft.Performance.SDK/Processing/ITableBuilder.cs
```diff
@@ -1,17 +1,52 @@
...
+        ///     This instance of the builder.
+        /// </returns>
+        /// <exception cref="ArgumentNullException">
+        ///     <paramref name="command"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="InvalidOperationException">
+        ///     A command with the same <see cref="TableCommand3.CommandName"/>
+        ///     has already been added to this instance (case-insensitive).
+        /// </exception>
+        ITableBuilder AddTableCommand3(TableCommand3 command);
```
xj-ms, line 41: Instead of adding a general `TableCommand3`, we should restrict it to only the table commands we support i.e.
```
ITableBuilder AddDownloadSourceCodeCommand(DownloadSourceCodeCommand command);
```

src/Microsoft.Performance.SDK/Processing/TableCommands/TableCommand3.cs
xj-ms: Another benefit of making it an interface is you can make implement it directly in the legacy classes like TableCommand2. No need for the standalone adapters

src/Microsoft.Performance.SDK/Processing/TableCommands/IColumnTableCommand.cs
```diff
@@ -0,0 +1,26 @@
...
+    ///     known statically via <see cref="ColumnId"/>, a host can exclude
+    ///     commands that are not relevant to a given column without invoking
+    ///     <see cref="TableCommand3{TContext, TResult}.CanExecute"/>.
+    /// </remarks>
+    public interface IColumnTableCommand
+    {
+        /// <summary>
+        ///     Gets the identifier of the column that this command targets.
+        /// </summary>
+        Guid ColumnId { get; }
```
xj-ms, line 24: I feel this should be an interface method `bool IsColumnSupported(IDataColumn)` method for `ITableCommand3`

src/Microsoft.Performance.SDK.Runtime/TableBuilder.cs
```diff
@@ -68,7 +69,14 @@ public TableBuilder()
...
 
-        public IReadOnlyList<TableCommand2> Commands { get; }
+        /// <summary>
+        ///     Gets the collection of commands registered against this table.
+        ///     Commands are stored as instances of the abstract
+        ///     <see cref="TableCommand3"/> base type; hosts should inspect the
+        ///     runtime type of each command to determine whether they recognize
+        ///     it. Unknown command types may be safely ignored.
+        /// </summary>
+        public IReadOnlyList<TableCommand3> Commands { get; }
```
xj-ms, line 79: I'd like to define an extension method that makes it easier for WPA to know what it needs to handle:
````
static void Visit(this ITableCommand3 command,
  Action<TableCommand2> visitCommand2,
  Action<DownloadSourceCodeCommand> visitDownloadSourceCommand,
  ...) // more in the future
````

src/Microsoft.Performance.Toolkit.Engine/RuntimeExecutionResults.cs
jmaxson-ms: I'm following the pattern that was used when adding TableCommand2 in order to avoid touching a deprecated property.

src/Microsoft.Performance.SDK/Processing/TableCommands/TableCommand3`2.cs
jmaxson-ms: forgot to remove these. thanks

src/Microsoft.Performance.SDK/Processing/ITableBuilder.cs
jmaxson-ms: I think I prefer the more general approach.
- we don't need to ever worry about a huge number of methods
- this allows for a host and a plugin to have the ability to add extra functionality that doesn't exist in the SDK

xj-ms: Our primary use case is for WPA though. I don't think we should allow people to add commands that WPA doesn't support. That's just misleading

xj-ms: IMHO if someone wants to develop a different host than WPA, they should fork this repo. This repo needs to first make sure things work with WPA before worrying about other hosts

src/Microsoft.Performance.SDK/Processing/TableCommands/TableCommand3.cs
jmaxson-ms: Jing, I think we should talk offline about this proposed change. I'm not sure that, when thinking about what the proposed change would entail, it would work as smoothly as it seems at an abstract level. (wrt to the second part of your comment specifically)

src/Microsoft.Performance.SDK/Processing/TableCommands/ITableCommand3Result.cs
jmaxson-ms: I believe this is leftover from a legacy implementation. Thanks for pointing it out.

src/Microsoft.Performance.SDK/Processing/ITableBuilder.cs
jmaxson-ms: This is a different philosophy than we had when creating this SDK. Happy to chat offline about it.

src/Microsoft.Performance.SDK/Processing/TableCommands/IColumnTableCommand.cs
jmaxson-ms: Let's chat offline

src/Microsoft.Performance.SDK/Processing/ITableBuilder.cs
jmaxson-ms, line 41: It doesn't work to return ITableBuilder here if we want ITableBuidlerWithRowCount to also use this.

src/Microsoft.Performance.SDK/Processing/TableCommands/IColumnTableCommand.cs
jmaxson-ms: Chatting offline, we decided on `bool IsColumnSupported(Guid)`

src/Microsoft.Performance.SDK/Processing/TableCommands/TableCommand3.cs
jmaxson-ms: We're making this an interface but leaving the adapters.

src/Microsoft.Performance.SDK/Processing/ITableBuilder.cs
jmaxson-ms: We've decided to go ahead and make these specific command methods.

src/Microsoft.Performance.SDK/Processing/TableCommands/IColumnTableCommand.cs
jmaxson-ms: Note that this ties table commands to one or more columns, but we believe this has been and will continue to be the case. If the need arises, we could re-evaluate in future changes.

src/Microsoft.Performance.SDK/Processing/ITableBuilder.cs
jmaxson-ms: Note that this does not tie the host to WPA, but it does restrict commands to those supported by the SDK.

PR #442, opened 2026-08-11
Populate a column's description and short description from each other

xj-ms:
All existing Xperf columns use either Description or ShortDescription. These 2 properties are basically used interchangeably in reality.
As a result, depending on which one is set, column descriptions are often only available either as tooltips or inside the table config dialog in WPA, but never both.

While plugin authors are responsible for providing both of them, one improvement we can do in SDK is to provide default values for them, so that they remain useful even if only one of them is set --- a short description can be the description, and a description can a short description after truncation. Plugin authors can still overwrite the default values if they want, and they are encouraged to do so

### DIFF
--- a/src/Microsoft.Performance.SDK/Processing/ColumnMetadata.cs
+++ b/src/Microsoft.Performance.SDK/Processing/ColumnMetadata.cs
@@ -226,7 +226,7 @@ namespace Microsoft.Performance.SDK.Processing
 
                 this.shortDescription = value;
 
-                // Overwite the description if the current value isn't informative
+                // Overwrite the description if the current value isn't informative
                 if (string.Equals(this.Description, this.Name, StringComparison.Ordinal))
                 {
                     this.Description = value;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Performance.SDK/Processing/ColumnMetadata.cs
```diff
@@ -196,14 +205,40 @@ public string Name
         public IProjection<int, string> NameProjection { get; }
 
         /// <summary>
-        /// Gets the user friendly short description of this column.
+        ///     Gets or sets a brief, user friendly description of this column,
+        ///     suitable for places with little room such as a tooltip.
+        ///     By default, this is initialized from <see cref="Description"/>, truncated if too long.
```
jmaxson-ms: we might want to provide a more precise value than `too long`. Unless a value is called out elsewhere and I just missed it? Even then, it might be good to add here.

xj-ms:
```suggestion
        ///     By default, this is initialized from <see cref="Description"/>, truncated if over 50 characters.
```

jmaxson-ms (approved):

malavshah11 (approved):

PR #445, opened 2026-09-10
Add support for column commands and download source code.

jmaxson-ms:
### New namespace: Microsoft.Performance.SDK.ColumnCommands
- DataColumnCommands — sealed container that carries the optional commands available on a column. Exposes:
- DataColumnCommands.Empty
- TryGetDownloadSourceCodeCommand(out DownloadSourceCodeCommand)
- DownloadSourceCodeCommand — abstract base a plugin implements to advertise "download source code" for a column value:
- string CommandName
- bool CanExecute(object value, string downloadPath)
- Task<DownloadSourceCodeResult> ExecuteAsync(object value, string downloadPath, CancellationToken)
- DownloadSourceCodeResult — result type with a success Uri (typically a local file URI the host opens) or a failure with ErrorMessage and optional attempted Uri.
New interface: IDataColumnCommands
- Implemented by DataColumn<T> and HierarchicalDataColumn<T> so hosts can retrieve the DataColumnCommands associated with any column.

## Column builder integration
- ColumnBuilder<T>.WithCommands(DataColumnCommands) — attach commands to the built column, mirroring the existing WithVariants(...) pattern.
- New WithToggle<T>(..., DataColumnCommands) and WithHierarchicalToggle<T>(..., DataColumnCommands) overloads on ToggleableColumnBuilder, plus equivalent overloads on ModalColumnBuilder and HierarchicalColumnBuilder<T>, so commands can also be scoped to a specific variant.
- DataColumn / HierarchicalDataColumn constructors accept an optional DataColumnCommands, and the runtime builders (EmptyColumnBuilder, ToggledColumnBuilder, ModalColumnWithModesBuilder) flow it through when materializing columns.

## Design notes
- DataColumnCommands is non-generic. The commands operate on object value because for hierarchical columns the row value may differ from T due to an ICollectionAccessProvider<T, TOut>. This is documented on the type.
- DownloadSourceCodeCommand itself is non-generic for the same reason and to keep the host-side dispatch simple.
- Commands were placed on the column (via IDataColumn / IHierarchicalDataColumn) rather than on ITableBuilder, because they are intrinsic column data (like ColumnConfiguration or the projection). The tradeoffs are captured inline in DataColumnCommands.cs.


src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnBuilder`1.cs
```diff
@@ -0,0 +1,186 @@
...
+    ///     Gets the projection that produces the column's values.
+    /// </summary>
+    protected IProjection<int, T> Projection { get; }
+
+    /// <summary>
+    ///     Gets or sets the delegate used to configure column variants on
+    ///     the built column, or <c>null</c> if no variants have been
+    ///     configured via <see cref="WithVariants"/>.
+    /// </summary>
+    protected Func<RootColumnBuilder, ColumnBuilder>? VariantOptions { get; set; } = null;
```
xj-ms: private set?

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnBuilder`1.cs
```diff
@@ -0,0 +1,186 @@
...
+    ///     column.
+    /// </summary>
+    /// <param name="tableBuilder">
+    ///     The table builder to which the built column is added.
+    /// </param>
+    /// <returns>
+    ///     The <paramref name="tableBuilder"/>, to allow chaining additional
+    ///     table-building calls.
+    /// </returns>
+    public ITableBuilderWithRowCount AddColumnToTable(
```
xj-ms: This breaks table builder's fluent syntax. TableBuilder should be able to chain into adding columns e.g.
```
tableBuilder.DoSomething(...).AddColumn(
  new ColumnBuilder(...).DoSomething(...)
)
.DoMoreThings()...
```

src/Microsoft.Performance.SDK.Runtime/ColumnBuilding/Builders/ToggledColumnBuilder.cs
```diff
@@ -58,6 +59,15 @@ internal override void Commit()
...
         IProjection<int, T> projection)
+    {
+        return WithToggle(toggleDescriptor, projection, default);
+    }
+
+    /// <inheritdoc />
+    public override ToggleableColumnBuilder WithToggle<T>(
+        ColumnVariantDescriptor toggleDescriptor,
+        IProjection<int, T> projection,
+        DataColumnCommands dataColumnCommands)
```
xj-ms, line 70: Maybe we should make the commands fluent too? So users can do things like 
```
variantBuilder.WithToggle(...,
  new Commands().WithSourceCommand(...).WithOtherCommands(...)
)
```

src/Microsoft.Performance.SDK/Processing/ColumnBuilding/ColumnBuilder`1.cs
jmaxson-ms: Good catch. Thanks!

src/Microsoft.Performance.SDK.Runtime/ColumnBuilding/Builders/ToggledColumnBuilder.cs
jmaxson-ms: I briefly considered that at first but seemed challenging because the existing implementation - they're not actually fluent as-is.

I'll take a longer look at this now that most of the other functionality is in place.

jmaxson-ms: Updated.

xj-ms:
It seems your local git user isn't correctly configured that the commits weren't created with jmaxson-ms. Might want to fix it in the long run

documentation/Using-the-SDK/Advanced/Adding-Column-Commands.md
```diff
@@ -102,12 +102,12 @@ tableBuilderWithRowCount.AddColumn(
     new DataColumn<Uri>(sourceColumnConfiguration, sourceProjection, commands));
 ```
 
-The strongly typed `ColumnBuilder<T>` provides an equivalent fluent form and can also configure variants:
+The strongly typed `ColumnBuilderBuilder<T>` can be used with `ITableBuilderWithRowCount` to add a column to a table:
```
xj-ms: typo: ColumnBuilderBuilder

xj-ms (approved):

helenkzhang (approved):

